PR #35, opened 2016-08-29
Much more strongly typed (and usable) file system interface.

zaq32:
* From now on all file system procedures receive pointer to FileSystem object as their first parameter to facilitate building abstraction layer over file system driver and allow much easier file system mocking,
* Separate operation status from context information returned by the
 * open,
 * read,
 * write
 * openDirectory
 procedures. From now on they returned dedicated types with separatied operation status (OSResult) and context information (i.e. fileHandle, number of bytes transfered) instead of pure ints,
* Migrate from pure int return values to OSResults,
* Map most of the errno error codes to OSResult enumerator,
* Update interface of open file procedure to:
 * separate file access mode (r, w, rw), from file opening mode,
 * drop the mode parameter that was supposed to be hard coded and will most likely never be changed.
 * introduce explicit enumerators for file access mode & file opening modes
* Update FsMock to support new file system interface.
* Fix doxygen issues.

---
This change is [Reviewable](https://reviewable.io/reviews/pw-sat2/pwsat2obc/35)

### DIFF
--- a/libs/base/Include/base/os.h
+++ b/libs/base/Include/base/os.h
@@ -30,6 +30,8 @@ typedef enum {
     OSResultTimeout = 2,
     /** @brief Requested operation is invalid. */
     OSResultInvalidOperation = 3,
+    /** @brief Requested element was not found. */
+    OSResultNotFound = 4,
 } OSResult;
 
 /**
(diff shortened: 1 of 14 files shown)


Novakov:
Reviewed 14 of 14 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/35)*

PR #61, opened 2016-12-14
N25Q flash driver

Novakov:
N25Q Flash memory driver and Yaffs driver

By default STK NAND flash is used. It is possible to enable N25Q instad by using `USE_EXTERNAL_FLASH=1` CMake flag,

Things to do in next pull requests:
* Tidy-up SPI driver
* Enable usage of many N25Q memories as different mount points

---
This change is [Reviewable](https://reviewable.io/reviews/pw-sat2/pwsat2obc/61)


zaq32:
Reviewed 15 of 57 files at r1.
Review status: 15 of 57 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[integration_tests/tests/base.py, line 28 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-KZJFQ88FsrIgXlMtpCQ:-KZJFQ88FsrIgXlMtpCR:bvk0zwo) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/integration_tests/tests/base.py#L28)):*
> ```Python
> 
>     def tearDown(self):
>         self.system.obc.sync_fs()
> ```

what is the purpose of this?

---

*[integration_tests/tests/test_antenna.py, line 15 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-KZJFJlYou75RVYXOdA3:-KZJFJlYou75RVYXOdA4:b-tv3ox5) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/integration_tests/tests/test_antenna.py#L15)):*
> ```Python
>         event = TestEvent()
>         self.system.primary_antenna.on_reset = event.set
>         self.power_on_and_wait()
> ```

why wait for obc startup?

---

*[integration_tests/tests/test_antenna.py, line 21 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-KZJFPB5KDvML9Y7yV2S:-KZJFPB5KDvML9Y7yV2T:b-tv3ox5) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/integration_tests/tests/test_antenna.py#L21)):*
> ```Python
>         event = TestEvent()
>         self.system.backup_antenna.on_reset = event.set
>         self.power_on_and_wait()
> ```

why wait for obc startup?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61)*

Novakov:
Review status: 15 of 57 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[integration_tests/tests/base.py, line 28 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-KZJFQ88FsrIgXlMtpCQ:-KZLXM-HAT7-r8EJub74:b68dfkx) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/integration_tests/tests/base.py#L28)):*

what is the purpose of this?
</blockquote></details>

Mounting file system can take long time (up to 30-40 seconds). Syncing file system forces YAFFS to write checkpoint information speeding up next mount significally (40sec -> 0.1sec)

---

*[integration_tests/tests/test_antenna.py, line 15 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-KZJFJlYou75RVYXOdA3:-KZLYEWj33dmEj4TWT8t:b-la1aar) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/integration_tests/tests/test_antenna.py#L15)):*

why wait for obc startup?
</blockquote></details>

If mouting flash will take longer that 1 second this test will fail

---

*[integration_tests/tests/test_antenna.py, line 21 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-KZJFPB5KDvML9Y7yV2S:-KZLYO7R3MrOwVd1zZWV:b-la1aar) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/integration_tests/tests/test_antenna.py#L21)):*

why wait for obc startup?
</blockquote></details>

If mouting flash will take longer that 1 second this test will fail

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61)*

zaq32:
Reviewed 1 of 57 files at r1.
Review status: 16 of 57 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61)*

Novakov:
[[Driver] Zewnętrzny FLASH](https://trello.com/c/nrpIgDKD/66-driver-zewnetrzny-flash)

zaq32:
Reviewed 41 of 57 files at r1.
Review status: all files reviewed at latest revision, 26 unresolved discussions, some commit checks failed.

---

*[libs/base/os_base.cpp, line 24 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0ZJ0usTIJdPdg_uGD:-K_0ZJ0usTIJdPdg_uGE:b-gxoe7v) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/libs/base/os_base.cpp#L24)):*
> ```cpp
> 
> 
> constexpr std::uint32_t MilisecondsToTicks(std::uint32_t miliseconds)
> ```

use TimeSpan type to represent time values. 

---

*[libs/base/os_base.cpp, line 29 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0fX5HusZIXbjewiZl:-K_0fX5HusZIXbjewiZm:b-9gi4dt) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/libs/base/os_base.cpp#L29)):*
> ```cpp
> }
> 
> Timeout::Timeout(std::uint32_t timeout) : _expireAt(System::GetTickCount() + MilisecondsToTicks(timeout))
> ```

I'd suggest measuring time in ms not system ticks, that way you would be able to drop the private dependency on freeRTOS from this module, which seems wrong anyway

---

*[libs/base/Include/system.h, line 56 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0Zy65UM0CWM2SKjwc:-K_0Zy65UM0CWM2SKjwd:bdsv8o0) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/libs/base/Include/system.h#L56)):*
> ```objc
> #define HAS_FLAG(value, flag) (has_flag(value, flag))
> 
> template <typename Enum> constexpr bool has_flag(Enum value, Enum flag)
> ```

mark it as inline

---

*[libs/base/Include/system.h, line 64 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0ZhpyM7YAXnJXdtNm:-K_0ZhpzyYaFrg-NpEcF:b5gkyww) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/libs/base/Include/system.h#L64)):*
> ```objc
> }
> 
> template <> constexpr bool has_flag(std::uint32_t value, std::uint32_t flag)
> ```

there is no need to specialize function template, c++ will be fine with just plain function overload.

---

*[libs/base/Include/system.h, line 64 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0ZzDamMyoPTMPVQyZ:-K_0ZzDamMyoPTMPVQy_:bdsv8o0) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/libs/base/Include/system.h#L64)):*
> ```objc
> }
> 
> template <> constexpr bool has_flag(std::uint32_t value, std::uint32_t flag)
> ```

mark it as inline

---

*[libs/base/Include/base/os.h, line 425 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0_1OuKA3h4saFNb5V:-K_0_1OuKA3h4saFNb5W:b-cpewpa) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/libs/base/Include/base/os.h#L425)):*
> ```objc
>      * @return Number of ticks
>      */
>     static std::uint32_t GetTickCount();
> ```

Number of ticks of what kind exactly? GetTickCount without a unit is completely useless. And since value combined unit is a time use type that is designed to represent time.

---

*[libs/base/Include/base/os.h, line 630 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0_UPkIIzimwKsGsVW:-K_0_UPkIIzimwKsGsVX:bqgkooy) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/libs/base/Include/base/os.h#L630)):*
> ```objc
>  * This class uses system tick count to measure elapsed time.
>  */
> class Timeout
> ```

What is exact designed use case for this class? 

---

*[libs/base/Include/base/os.h, line 637 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0_QPSNrPNysqZ_h0o:-K_0_QPSNrPNysqZ_h0p:bf883yv) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/libs/base/Include/base/os.h#L637)):*
> ```objc
>      * @param[in] timeout Timeout in miliseconds
>      */
>     Timeout(std::uint32_t timeout);
> ```

Use TimeSpan instead

---

*[libs/drivers/n25q/n25q.cpp, line 51 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0a2bnHrvNzoyjn8cp:-K_0a2bnHrvNzoyjn8cq:bea5o78) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/libs/drivers/n25q/n25q.cpp#L51)):*
> ```cpp
>     ReaderInitialize(&reader, response.data(), 3);
> 
>     Id id;
> ```

it would be better to initialize objects via constructors, this way it will be much harder to omit some of its fields.

---

*[libs/drivers/n25q/n25q.cpp, line 144 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0awyaxlvtgOk387d-:-K_0awyaxlvtgOk387d0:b65evxv) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/libs/drivers/n25q/n25q.cpp#L144)):*
> ```cpp
> }
> 
> bool N25QDriver::WaitBusy(std::uint32_t timeout)
> ```

if there is no other way I can accept active register polling, but I would recommend at least sleeping for 0-1 ms (exact value TBD) during the iterations or at least yielding the task control so one task will not over-zealously block tasks with lover priority while not doing anything useful.

---

*[libs/drivers/n25q/Include/n25q/n25q.h, line 23 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0d2JYLZMLHYiW02Dx:-K_0d2JYLZMLHYiW02Dy:byslh8m) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/libs/drivers/n25q/Include/n25q/n25q.h#L23)):*
> ```objc
>          * @brief Device ID structure
>          */
>         struct Id
> ```

raw struct like this may be useful but it is very easy to accidentally left it uninitialized. Add at least:
- default constructor
- parametrized constructor that initializes all fields.

---

*[libs/drivers/n25q/Include/n25q/n25q.h, line 81 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0g44Q25FgjAYKg83j:-K_0g44Q25FgjAYKg83k:b-hbnfm6) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/libs/drivers/n25q/Include/n25q/n25q.h#L81)):*
> ```objc
>          *  Timeouts are based on maximum times specified in datasheet + some margin of error (20%)
>          */
>         class N25QDriver final
> ```

Since this class is not internally synchronized am I correct in assuming that high level driver (in this case yaffs) is resposnible for doring so?

---

*[libs/drivers/n25q/Include/n25q/n25q.h, line 81 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0gPizYwRPxbnMgKGg:-K_0gPizYwRPxbnMgKGh:b-m561ui) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/libs/drivers/n25q/Include/n25q/n25q.h#L81)):*
> ```objc
>          *  Timeouts are based on maximum times specified in datasheet + some margin of error (20%)
>          */
>         class N25QDriver final
> ```

I would not mind having the operation timeout specified for each function in this driver as they are seem to quite big...

---

*[libs/drivers/n25q/Include/n25q/n25q.h, line 144 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0cm-tHHW0ugjk7E3t:-K_0cm-tHHW0ugjk7E3u:b19bk24) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/libs/drivers/n25q/Include/n25q/n25q.h#L144)):*
> ```objc
>              * This operating can take up to 4 minutes, so be patient
>              */
>             OperationResult EraseChip();
> ```

If it really takes this long than it should be asynchronous.

---

*[libs/drivers/n25q/Include/n25q/n25q.h, line 159 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0cvdgCk0N3orM-FO6:-K_0cvdgCk0N3orM-FO7:b-cyq1x6) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/libs/drivers/n25q/Include/n25q/n25q.h#L159)):*
> ```objc
>              * @return true of operation finished, false on timeout
>              */
>             bool WaitBusy(std::uint32_t timeout);
> ```

use TimeSpan instead of raw integer

---

*[libs/drivers/n25q/Include/n25q/yaffs.h, line 137 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0hbvRpEyfWhdwvmZH:-K_0hbvRpEyfWhdwvmZI:b2eey6f) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/libs/drivers/n25q/Include/n25q/yaffs.h#L137)):*
> ```objc
> 
>         /**
>          * @brief Helper class allowing to specify @ref N25QYaffsDeviceBase parameters as template arguments
> ```

what is actually the point of this? These template parameters are immediately converted to runtime parameters that can be overridden, which makes them illusion at best.

---

*[libs/drivers/spi/efm.cpp, line 11 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0lcgDLMTDpghD5AgX:-K_0lcgDLMTDpghD5AgY:b-qcaz01) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/libs/drivers/spi/efm.cpp#L11)):*
> ```cpp
> void EFMSPIInterface::Select()
> {
>     GPIO_PinOutClear(gpioPortD, 3);
> ```

Shouldn't this be specified in io_map.h?

---

*[libs/drivers/spi/Include/spi/efm.h, line 24 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0lnP5JzEVHxXJ0ROb:-K_0lnP5JzEVHxXJ0ROc:bn9j0rp) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/libs/drivers/spi/Include/spi/efm.h#L24)):*
> ```objc
>          * @brief SPI interfaced using SPIDRV
>          */
>         class EFMSPIInterface final : public ISPIInterface
> ```

Who is responsible for ensuring that access to the spi is synchronized?

---

*[libs/drivers/spi/Include/spi/spi.h, line 64 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0dnSvpbVvtOsV94bi:-K_0dnSvpbVvtOsV94bj:bqqwhia) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/libs/drivers/spi/Include/spi/spi.h#L64)):*
> ```objc
>             SPISelectSlave& operator=(const SPISelectSlave&) = delete;
>             SPISelectSlave(SPISelectSlave&&) = delete;
>             SPISelectSlave& operator=(SPISelectSlave&&) = delete;
> ```

move constructor & move assignment operator can be safely & correctly defined I would recommend defining those.

---

*[libs/fs/fs.cpp, line 160 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0ePQQ9OIDrL6_eE0U:-K_0ePQQ9OIDrL6_eE0V:b-6rmzci) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/libs/fs/fs.cpp#L160)):*
> ```cpp
>  * @param root Root directory
>  */
> static void RemoveDirectoryContents(yaffs_obj* root)
> ```

I'd limit the number of possible nested calls (due to uncertain stack requirements) and in such case fallback to full driver erase.

---

*[libs/yaffs_glue/yaffs_os_glue.cpp, line 16 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0exVrhKQ6OaQUOTjh:-K_0exVrhKQ6OaQUOTji:b-9n77cj) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/libs/yaffs_glue/yaffs_os_glue.cpp#L16)):*
> ```cpp
> int yaffsError = 0;
> unsigned int yaffs_trace_mask =
>     YAFFS_TRACE_ERASE | YAFFS_TRACE_ERROR | YAFFS_TRACE_BUG | YAFFS_TRACE_BAD_BLOCKS | YAFFS_TRACE_BUFFERS | YAFFS_TRACE_MOUNT;
> ```

do we need full tracing in release builds?

---

*[src/storage.h, line 33 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0jz4oQlPMTN-2XW8L:-K_0jz4oQlPMTN-2XW8M:bngzeeb) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/src/storage.h#L33)):*
> ```objc
>     void Initialize();
> 
> #ifdef USE_EXTERNAL_FLASH
> ```

Once or twice conditional compilation is acceptable but this symbol is used in quite a few places (here, this class methods and even main & obc commands). I would recommend moving the variable parts to two independent libraries and conditionally link them in instead of spreading #ifdefs over the code

---

*[src/commands/file_system.cpp, line 10 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0kVpdIU6cXxgrjakB:-K_0kVpeRuLLfbqKA32S:b-lxo260) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/src/commands/file_system.cpp#L10)):*
> ```cpp
> #include "yaffs_guts.h"
> #include "yaffsfs.h"
> }
> ```

provide dedicated header i.e yaffs.hpp that wold handle interfacing C with C++ 

---

*[unit_tests/SPI/SPIMock.h, line 32 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0j6h4h5F-sXOran0l:-K_0j6h4h5F-sXOran0m:b-wzrry8) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/unit_tests/SPI/SPIMock.h#L32)):*
> ```objc
> };
> 
> SPIExpectSelected::SPIExpectSelected(SPIInterfaceMock& mock) : _mock(mock)
> ```

this function should be marked as inline, however as a rule constructors almost always should not be defined inline.

---

*[unit_tests/SPI/SPIMock.h, line 37 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0j6HkF2THcXpr_Tbn:-K_0j6HkF2THcXpr_Tbo:b7gsl2j) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/unit_tests/SPI/SPIMock.h#L37)):*
> ```objc
> }
> 
> SPIExpectSelected::~SPIExpectSelected()
> ```

this function should be marked as inline, however as a rule destructors almost always should not be defined inline.

---

*[unit_tests/SPI/SPIMock.h, line 42 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0j0msCZFCiLoPRIiK:-K_0j0msCZFCiLoPRIiL:bgj81gu) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/unit_tests/SPI/SPIMock.h#L42)):*
> ```objc
> }
> 
> SPIExpectSelected SPIInterfaceMock::ExpectSelected()
> ```

this function should be marked as inline

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61)*

Novakov:
Review status: all files reviewed at latest revision, 26 unresolved discussions, some commit checks failed.

---

*[libs/base/os_base.cpp, line 29 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0fX5HusZIXbjewiZl:-K_0pvje_HtEiZr-fwY1:bt3hyt2) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/libs/base/os_base.cpp#L29)):*

I'd suggest measuring time in ms not system ticks, that way you would be able to drop the private dependency on freeRTOS from this module, which seems wrong anyway
</blockquote></details>

It would be nice but currently our timer is not-moving forward unless kicked by terminal and I wanted real timeouts for flash-related operations. 

---

*[libs/base/Include/system.h, line 64 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0ZzDamMyoPTMPVQyZ:-K_0qSp276opE6LS_htK:b6ne606) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/libs/base/Include/system.h#L64)):*
According to http://stackoverflow.com/questions/14391272/does-constexpr-imply-inline

> Yes (§7.1.5/2): "constexpr functions and constexpr constructors are implicitly inline (7.1.2)."

---

*[libs/drivers/n25q/Include/n25q/n25q.h, line 81 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0g44Q25FgjAYKg83j:-K_0oAAG7Ib6XZ10foO2:b2nhswy) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/libs/drivers/n25q/Include/n25q/n25q.h#L81)):*

Since this class is not internally synchronized am I correct in assuming that high level driver (in this case yaffs) is resposnible for doring so?
</blockquote></details>

Yes

---

*[libs/drivers/n25q/Include/n25q/n25q.h, line 81 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0gPizYwRPxbnMgKGg:-K_0oAygyrflxOUXmhtE:b-32jzlt) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/libs/drivers/n25q/Include/n25q/n25q.h#L81)):*

I would not mind having the operation timeout specified for each function in this driver as they are seem to quite big...
</blockquote></details>

There are defined (and checked) timeouts for each function (except read as they cannot timeout).

If you would like to get timeout parameter for each function in what non-default values will be used?

---

*[libs/drivers/n25q/Include/n25q/n25q.h, line 144 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0cm-tHHW0ugjk7E3t:-K_0oeArki7I95ohGDyS:b-crlfav) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/libs/drivers/n25q/Include/n25q/n25q.h#L144)):*

If it really takes this long than it should be asynchronous.
</blockquote></details>

If you look closely, you will see that only way to invoke this method is via terminal in which you really want to wait for this operation to end.

---

*[libs/drivers/n25q/Include/n25q/yaffs.h, line 137 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0hbvRpEyfWhdwvmZH:-K_0oscPhTlTUVCY0nME:b-p8vi9u) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/libs/drivers/n25q/Include/n25q/yaffs.h#L137)):*

what is actually the point of this? These template parameters are immediately converted to runtime parameters that can be overridden, which makes them illusion at best.
</blockquote></details>

The only reason is possiblity to see flash parameters just by looking through `OBCStorage` class without need to dig into it's constructor (just like with tasks)

---

*[libs/drivers/spi/efm.cpp, line 11 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0lcgDLMTDpghD5AgX:-K_0pAKt8GqL4oQfU3ZR:b21b85h) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/libs/drivers/spi/efm.cpp#L11)):*

Shouldn't this be specified in io_map.h?
</blockquote></details>

More reasonable SPI driver is in-progress (see Trello)

---

*[libs/drivers/spi/Include/spi/efm.h, line 24 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0lnP5JzEVHxXJ0ROb:-K_0pI6YSPs2s1DZi0HM:b21b85h) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/libs/drivers/spi/Include/spi/efm.h#L24)):*

Who is responsible for ensuring that access to the spi is synchronized?
</blockquote></details>

More reasonable SPI driver is in-progress (see Trello)

---

*[libs/drivers/spi/Include/spi/spi.h, line 64 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0dnSvpbVvtOsV94bi:-K_0pIyTrG1WT88qRpJZ:bx375n3) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/libs/drivers/spi/Include/spi/spi.h#L64)):*

move constructor & move assignment operator can be safely & correctly defined I would recommend defining those.
</blockquote></details>

Honestly, I don't see any reason to pass selected slave around. Making it impossible saves us from possible trouble of debugging when slave is selected/deselected 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61)*

zaq32:
Review status: all files reviewed at latest revision, 22 unresolved discussions, some commit checks failed.

---

*[libs/base/os_base.cpp, line 29 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0fX5HusZIXbjewiZl:-K_0qulmeI9nz1IWhd0K:b-a7uzbw) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/libs/base/os_base.cpp#L29)):*

It would be nice but currently our timer is not-moving forward unless kicked by terminal and I wanted real timeouts for flash-related operations. 
</blockquote></details>

This has nothing to do with the timer. TimeSpan type is conceptually independent of *any timer* and can be used to measure time flow in general.

---

*[libs/drivers/n25q/Include/n25q/n25q.h, line 81 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0gPizYwRPxbnMgKGg:-K_0rMKVG_FqWs3DhcmS:b-fn51h0) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/libs/drivers/n25q/Include/n25q/n25q.h#L81)):*

There are defined (and checked) timeouts for each function (except read as they cannot timeout).

If you would like to get timeout parameter for each function in what non-default values will be used?
</blockquote></details>

I would like to have timeouts explicitly _documented_ for each potentially long running (>= 0.5 ms)  operation.

---

*[libs/drivers/n25q/Include/n25q/n25q.h, line 144 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0cm-tHHW0ugjk7E3t:-K_0rdWi1l6cpIUWWp3L:bz1hvfy) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/libs/drivers/n25q/Include/n25q/n25q.h#L144)):*

If you look closely, you will see that only way to invoke this method is via terminal in which you really want to wait for this operation to end.
</blockquote></details>

for the moment yes, can you guarantee that it will stay this way?

---

*[libs/drivers/n25q/Include/n25q/yaffs.h, line 137 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0hbvRpEyfWhdwvmZH:-K_0rrJ4oJ3G7mDXLEH7:b-tlj2mx) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/libs/drivers/n25q/Include/n25q/yaffs.h#L137)):*

The only reason is possiblity to see flash parameters just by looking through `OBCStorage` class without need to dig into it's constructor (just like with tasks)
</blockquote></details>

So I wold recommend using those values in such way that they cannot be overwritten at runtime

---

*[libs/drivers/spi/Include/spi/spi.h, line 64 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0dnSvpbVvtOsV94bi:-K_0sFTpgtPBguPaOKsL:b-sbv38x) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/libs/drivers/spi/Include/spi/spi.h#L64)):*

Honestly, I don't see any reason to pass selected slave around. Making it impossible saves us from possible trouble of debugging when slave is selected/deselected 
</blockquote></details>

I was thinking of something like this:
```
SPISelectSlave Prepare();

/* ... */
{
   auto select = Prepare()
   // finish the operation
}
```

In essence this will allow for easy extension of the chip selection beyond the lifetime of single function call.

However this is only a suggestion...

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61)*

Novakov:
Review status: 31 of 67 files reviewed at latest revision, 21 unresolved discussions.

---

*[libs/base/os_base.cpp, line 24 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0ZJ0usTIJdPdg_uGD:-K_9y4AHDnQUdgmE2TrZ:b-896fix) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/libs/base/os_base.cpp#L24)):*

use TimeSpan type to represent time values. 
</blockquote></details>

Done.

---

*[libs/base/os_base.cpp, line 29 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0fX5HusZIXbjewiZl:-K_9y6R2pbramtJYaLXQ:b-59msik) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/libs/base/os_base.cpp#L29)):*

This has nothing to do with the timer. TimeSpan type is conceptually independent of *any timer* and can be used to measure time flow in general.
</blockquote></details>

Right, now I got it :)

---

*[libs/base/Include/system.h, line 64 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0ZhpyM7YAXnJXdtNm:-K_9jqPsAHbWE7fO6b0D:b4690ga) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/libs/base/Include/system.h#L64)):*

there is no need to specialize function template, c++ will be fine with just plain function overload.
</blockquote></details>

Unless you want to use this function with value that contains flags but is not enum (like peripheral register)

---

*[libs/base/Include/base/os.h, line 425 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0_1OuKA3h4saFNb5V:-K_9yMT9pYBc6MtiqLPs:b-896fix) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/libs/base/Include/base/os.h#L425)):*

Number of ticks of what kind exactly? GetTickCount without a unit is completely useless. And since value combined unit is a time use type that is designed to represent time.
</blockquote></details>

Done.

---

*[libs/base/Include/base/os.h, line 630 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0_UPkIIzimwKsGsVW:-K_9ypv5oY9aCmrhuLLY:b-9uug86) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/libs/base/Include/base/os.h#L630)):*

What is exact designed use case for this class? 
</blockquote></details>

Added to documentation

---

*[libs/base/Include/base/os.h, line 637 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0_QPSNrPNysqZ_h0o:-K_9yOJZT0gtfrVmhJt0:b-896fix) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/libs/base/Include/base/os.h#L637)):*

Use TimeSpan instead
</blockquote></details>

Done.

---

*[libs/drivers/n25q/n25q.cpp, line 51 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0a2bnHrvNzoyjn8cp:-K_9kpgXT5rGgmKj5MCX:b-896fix) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/libs/drivers/n25q/n25q.cpp#L51)):*

it would be better to initialize objects via constructors, this way it will be much harder to omit some of its fields.
</blockquote></details>

Done.

---

*[libs/drivers/n25q/n25q.cpp, line 144 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0awyaxlvtgOk387d-:-K_9lQDtUGAvGkH0qf61:b-896fix) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/libs/drivers/n25q/n25q.cpp#L144)):*

if there is no other way I can accept active register polling, but I would recommend at least sleeping for 0-1 ms (exact value TBD) during the iterations or at least yielding the task control so one task will not over-zealously block tasks with lover priority while not doing anything useful.
</blockquote></details>

Done.

---

*[libs/drivers/n25q/Include/n25q/n25q.h, line 23 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0d2JYLZMLHYiW02Dx:-K_9lgwIZoFMsBqskNPo:b-896fix) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/libs/drivers/n25q/Include/n25q/n25q.h#L23)):*

raw struct like this may be useful but it is very easy to accidentally left it uninitialized. Add at least:
- default constructor
- parametrized constructor that initializes all fields.
</blockquote></details>

Done.

---

*[libs/drivers/n25q/Include/n25q/n25q.h, line 81 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0gPizYwRPxbnMgKGg:-K_9pF5S0fpdWz5rMgnq:b-896fix) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/libs/drivers/n25q/Include/n25q/n25q.h#L81)):*

I would like to have timeouts explicitly _documented_ for each potentially long running (>= 0.5 ms)  operation.
</blockquote></details>

Done.

---

*[libs/drivers/n25q/Include/n25q/n25q.h, line 144 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0cm-tHHW0ugjk7E3t:-K_AmRvRtoju0iBHx95Y:b-r2xu4i) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/libs/drivers/n25q/Include/n25q/n25q.h#L144)):*

for the moment yes, can you guarantee that it will stay this way?
</blockquote></details>

Total number of things I can guarantee = 0. 

I can't see any reason (at least for now) and reasonable way to do it asynchronously. Doing so will require creating another task that will handle waiting for operation end - it can be created upfront (wasting memory) or on demand (harder memory calculations).

---

*[libs/drivers/n25q/Include/n25q/n25q.h, line 159 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0cvdgCk0N3orM-FO6:-K_A0uDmV2A-N4bihxzs:b-896fix) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/libs/drivers/n25q/Include/n25q/n25q.h#L159)):*

use TimeSpan instead of raw integer
</blockquote></details>

Done.

---

*[libs/drivers/n25q/Include/n25q/yaffs.h, line 137 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0hbvRpEyfWhdwvmZH:-K_9pmwK1IH-2BxbBsaF:b-896fix) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/libs/drivers/n25q/Include/n25q/yaffs.h#L137)):*

So I wold recommend using those values in such way that they cannot be overwritten at runtime
</blockquote></details>

Done.

---

*[libs/drivers/spi/Include/spi/spi.h, line 64 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0dnSvpbVvtOsV94bi:-K_AnsGXwMZ4lZ1pBNhz:bdb42u1) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/libs/drivers/spi/Include/spi/spi.h#L64)):*

I was thinking of something like this:
```
SPISelectSlave Prepare();

/* ... */
{
   auto select = Prepare()
   // finish the operation
}
```

In essence this will allow for easy extension of the chip selection beyond the lifetime of single function call.

However this is only a suggestion...
</blockquote></details>

It's nice feature to have if it will be needed to pass selected slave from function to function.

---

*[libs/fs/fs.cpp, line 160 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0ePQQ9OIDrL6_eE0U:-K_BCqkCQiDY3M-kzKov:b-896fix) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/libs/fs/fs.cpp#L160)):*

I'd limit the number of possible nested calls (due to uncertain stack requirements) and in such case fallback to full driver erase.
</blockquote></details>

Done.

---

*[libs/yaffs_glue/yaffs_os_glue.cpp, line 16 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0exVrhKQ6OaQUOTjh:-K_9q79_wLYoYEKvXrK-:b-bexbyc) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/libs/yaffs_glue/yaffs_os_glue.cpp#L16)):*

do we need full tracing in release builds?
</blockquote></details>

It is far from full trace ;) These flags allow to spot and diagnose the most common problems with yaffs when they occur. 

---

*[src/commands/file_system.cpp, line 10 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0kVpdIU6cXxgrjakB:-K_A2safx2bDal72e9TG:b-896fix) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/src/commands/file_system.cpp#L10)):*

provide dedicated header i.e yaffs.hpp that wold handle interfacing C with C++ 
</blockquote></details>

Done.

---

*[unit_tests/SPI/SPIMock.h, line 32 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0j6h4h5F-sXOran0l:-K_9tRl6tfcos9zV-5jp:b-896fix) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/unit_tests/SPI/SPIMock.h#L32)):*

this function should be marked as inline, however as a rule constructors almost always should not be defined inline.
</blockquote></details>

Done.

---

*[unit_tests/SPI/SPIMock.h, line 37 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0j6HkF2THcXpr_Tbn:-K_9tSQX08m9NLeQ0Pzg:b-896fix) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/unit_tests/SPI/SPIMock.h#L37)):*

this function should be marked as inline, however as a rule destructors almost always should not be defined inline.
</blockquote></details>

Done.

---

*[unit_tests/SPI/SPIMock.h, line 42 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0j0msCZFCiLoPRIiK:-K_9tTJIBwYQ8sX6gAcz:b-896fix) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/unit_tests/SPI/SPIMock.h#L42)):*

this function should be marked as inline
</blockquote></details>

Done.

---

*[src/storage.h, line 33 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0jz4oQlPMTN-2XW8L:-K_B5QoXWlaRC4QGll3Y:b-896fix) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/src/storage.h#L33)):*

Once or twice conditional compilation is acceptable but this symbol is used in quite a few places (here, this class methods and even main & obc commands). I would recommend moving the variable parts to two independent libraries and conditionally link them in instead of spreading #ifdefs over the code
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61)*

zaq32:
Reviewed 2 of 14 files at r2, 35 of 37 files at r3.
Review status: 66 of 67 files reviewed at latest revision, 8 unresolved discussions.

---

*[libs/base/Include/system.h, line 64 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0ZhpyM7YAXnJXdtNm:-K_EA14xo4ojODJK6ToL:b-az9awg) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/libs/base/Include/system.h#L64)):*

Unless you want to use this function with value that contains flags but is not enum (like peripheral register)
</blockquote></details>

Can you expand on this?

---

*[libs/base/Include/base/os.h, line 425 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0_1OuKA3h4saFNb5V:-K_E17PX1KBqdxjlNliK:bciipj3) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/libs/base/Include/base/os.h#L425)):*

Done.
</blockquote></details>

I rather thought of moving the TimeSpan type from time library to base so OS layer can use it.  This way it still is integer with no semantic meaning.

---

*[libs/base/Include/base/os.h, line 637 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0_QPSNrPNysqZ_h0o:-K_E18N6v4bExQgQdC9W:b-34vmyd) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/libs/base/Include/base/os.h#L637)):*

Done.
</blockquote></details>

See above

---

*[libs/drivers/n25q/Include/n25q/n25q.h, line 26 at r3](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_E1XbUUWd9StcJ8neN:-K_E1XbUUWd9StcJ8neO:b-u1kid6) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/552e532ba40626aa9f089ffd6d3c68278381fb57/libs/drivers/n25q/Include/n25q/n25q.h#L26)):*
> ```objc
>         struct Id
>         {
>             Id(std::uint8_t manufacturer, std::uint8_t memoryType, std::uint8_t memoryCapacity)
> ```

In general constructors  should not be defined inline 

---

*[libs/drivers/n25q/Include/n25q/yaffs.h, line 61 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_E2FqkH-UfWI_4lWMT:-K_E2FqkH-UfWI_4lWMU:bs29sx2) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/libs/drivers/n25q/Include/n25q/yaffs.h#L61)):*
> ```objc
>             N25QYaffsDeviceBase& operator=(const N25QYaffsDeviceBase&) = delete;
>             N25QYaffsDeviceBase(N25QYaffsDeviceBase&&) = delete;
>             N25QYaffsDeviceBase& operator=(N25QYaffsDeviceBase&&) = delete;
> ```

this will of course work but it would be much more readable if you put those declarations in a structs that  

```
struct NotCopyable
{
    NotCopyable(const NotCopyable& arg) = delete;
    NotCopyAble& operator=(const NotCopyable& arg) = delete;
};

struct NotMoveable
{
    NotMoveable(NotMoveable&& arg) = delete;
    NotMoveable& operator=(NotMoveable&& arg) = delete;
};
```

And simply private inherit from those.

---

*[libs/obc/storage/Include/obc/storage.h, line 43 at r3](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_E4PLpUTlygt2TrKDx:-K_E4PLqTbNE_0krPy1H:b-hhj4xm) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/552e532ba40626aa9f089ffd6d3c68278381fb57/libs/obc/storage/Include/obc/storage.h#L43)):*
> ```objc
>         OSResult ClearStorage();
> 
>         /** @brief Performs (potentally lengthy) erase operation */
> ```

potent**i**ally

---

*[libs/obc/storage/Include/obc/storage.h, line 46 at r3](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_E5GJTBPuRSOY_QIul:-K_E5GJTBPuRSOY_QIum:b-jmo2dp) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/552e532ba40626aa9f089ffd6d3c68278381fb57/libs/obc/storage/Include/obc/storage.h#L46)):*
> ```objc
>         OSResult Erase();
> 
> #ifdef USE_EXTERNAL_FLASH
> ```

Since external access to those fields is not desirable I would recommend putting them in private section.

---

*[libs/obc/storage/n25q_flash/CMakeLists.txt, line 10 at r3](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_E5vsuE9N09qi3fbvZ:-K_E5vsuE9N09qi3fbv_:b8ry2oq) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/552e532ba40626aa9f089ffd6d3c68278381fb57/libs/obc/storage/n25q_flash/CMakeLists.txt#L10)):*
> ```CMake
> 
> target_link_libraries(${NAME} 
> 	obc_storage
> ```

This is interesting. We have obc_storage library that is linked to this library & at the same time this library is linked to obc storage. As a result we have logical circular dependency something that should not have happened. 

I recommend doing something like this:
* dropp the obc_storage_* -> obc_storage dependency
* Make the OBCStorage class a template that is instantiated on different specific driver type based on build on build option:
to do that you would have to change the class OBCStorage into something like this:
```
 template <typename MemDriver> class OBCStorageHandler final
{
public:
    OBCStorageHandler (drivers::spi::ISPIInterface& spi, FileSystem& fs) :
        driver(spi, fs)
    {
    }

    void Initialize()
    {
        this->driver.Initialize();
    }

    OSResult ClearStorage()
    {
        return this->driver.ClearStorage();
    }
    
    OSResult Erase()
    {
        return this->driver.Erase();
    }
    
private:
    MemDriver driver;

};
```

and then finally:
```
#ifdef USE_EXTERNAL_FLASH
    #include "n25q.hpp"
    
    typedef OBCStorage<N25QProvider> OBCStorage;
#else
    #include "nand.h"
    
    typedef OBCStorage<NANDProvider> OBCStorage;
#endif
```

This way you have single common interface without the convoluted dependency graph. However this is a suggestion and not the only solution to this situation....

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61)*

Novakov:
Review status: 57 of 71 files reviewed at latest revision, 8 unresolved discussions.

---

*[libs/base/Include/system.h, line 64 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0ZhpyM7YAXnJXdtNm:-K_ERWeRaDaJsJsGg3A6:b-nrgthq) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/libs/base/Include/system.h#L64)):*

Can you expand on this?
</blockquote></details>

Type trait `std::underlying_type_t<Enum>` fails when template parameter is not enum which is case when using it with registers (see libs/drivers/i2c/efm.cpp, line 50)

---

*[libs/base/Include/base/os.h, line 425 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0_1OuKA3h4saFNb5V:-K_ETcKp7P2DT99_1bQe:b-e1syfb) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/libs/base/Include/base/os.h#L425)):*

I rather thought of moving the TimeSpan type from time library to base so OS layer can use it.  This way it still is integer with no semantic meaning.
</blockquote></details>

Good point, howver @mandrek44 will be working on incorporating `std::chrono` soon, so changing anything in `libs/time` will make harder merges.

---

*[libs/base/Include/base/os.h, line 637 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0_QPSNrPNysqZ_h0o:-K_EUzds_7KWkYlhtuuO:b-4pvm2l) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/libs/base/Include/base/os.h#L637)):*

See above
</blockquote></details>

See above ;)

---

*[libs/drivers/n25q/Include/n25q/n25q.h, line 26 at r3](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_E1XbUUWd9StcJ8neN:-K_EVOFZYsIDPz0LR-h0:b-896fix) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/552e532ba40626aa9f089ffd6d3c68278381fb57/libs/drivers/n25q/Include/n25q/n25q.h#L26)):*

In general constructors  should not be defined inline 
</blockquote></details>

Done.

---

*[libs/drivers/n25q/Include/n25q/yaffs.h, line 61 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_E2FqkH-UfWI_4lWMT:-K_ENb1dt2SXJJ-1YGKu:b2npxp1) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/libs/drivers/n25q/Include/n25q/yaffs.h#L61)):*

this will of course work but it would be much more readable if you put those declarations in a structs that  

```
struct NotCopyable
{
    NotCopyable(const NotCopyable& arg) = delete;
    NotCopyAble& operator=(const NotCopyable& arg) = delete;
};

struct NotMoveable
{
    NotMoveable(NotMoveable&& arg) = delete;
    NotMoveable& operator=(NotMoveable&& arg) = delete;
};
```

And simply private inherit from those.


</blockquote></details>

Nice :) Done

---

*[libs/obc/storage/Include/obc/storage.h, line 43 at r3](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_E4PLpUTlygt2TrKDx:-K_ENqZneKX7PI5d6oMB:b-896fix) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/552e532ba40626aa9f089ffd6d3c68278381fb57/libs/obc/storage/Include/obc/storage.h#L43)):*

potent**i**ally
</blockquote></details>

Done.

---

*[libs/obc/storage/Include/obc/storage.h, line 46 at r3](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_E5GJTBPuRSOY_QIul:-K_ENshmWkiJ8JEzejS8:b-896fix) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/552e532ba40626aa9f089ffd6d3c68278381fb57/libs/obc/storage/Include/obc/storage.h#L46)):*

Since external access to those fields is not desirable I would recommend putting them in private section.
</blockquote></details>

Done.

---

*[libs/obc/storage/n25q_flash/CMakeLists.txt, line 10 at r3](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_E5vsuE9N09qi3fbvZ:-K_F0nxJoEl6qwD4De4D:b-896fix) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/552e532ba40626aa9f089ffd6d3c68278381fb57/libs/obc/storage/n25q_flash/CMakeLists.txt#L10)):*

This is interesting. We have obc_storage library that is linked to this library & at the same time this library is linked to obc storage. As a result we have logical circular dependency something that should not have happened. 

I recommend doing something like this:
* dropp the obc_storage_* -> obc_storage dependency
* Make the OBCStorage class a template that is instantiated on different specific driver type based on build on build option:
to do that you would have to change the class OBCStorage into something like this:
```
 template <typename MemDriver> class OBCStorageHandler final
{
public:
    OBCStorageHandler (drivers::spi::ISPIInterface& spi, FileSystem& fs) :
        driver(spi, fs)
    {
    }

    void Initialize()
    {
        this->driver.Initialize();
    }

    OSResult ClearStorage()
    {
        return this->driver.ClearStorage();
    }
    
    OSResult Erase()
    {
        return this->driver.Erase();
    }
    
private:
    MemDriver driver;

};
```

and then finally:
```
#ifdef USE_EXTERNAL_FLASH
    #include "n25q.hpp"
    
    typedef OBCStorage<N25QProvider> OBCStorage;
#else
    #include "nand.h"
    
    typedef OBCStorage<NANDProvider> OBCStorage;
#endif
```

This way you have single common interface without the convoluted dependency graph. However this is a suggestion and not the only solution to this situation....


</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61)*

zaq32:
Reviewed 1 of 58 files at r1, 1 of 39 files at r3, 15 of 15 files at r4.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[libs/base/Include/system.h, line 64 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-K_0ZhpyM7YAXnJXdtNm:-K_F8yIPaY-qi3-f_cOi:bvyyxfr) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/42cd3c719f3531c2adc25d8f274529e5ed2dcf88/libs/base/Include/system.h#L64)):*

Type trait `std::underlying_type_t<Enum>` fails when template parameter is not enum which is case when using it with registers (see libs/drivers/i2c/efm.cpp, line 50)
</blockquote></details>

this is true but my point here is that you do not need template <> in front of this function declaration. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61)*

zaq32:
Reviewed 1 of 58 files at r1, 1 of 1 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61)*

zaq32:
:lgtm:

---

Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/61#-:-K_FIr2_MOejhdPyX01P:bnfp4nl)*

PR #62, opened 2016-12-18
Comm get exact number of bytes from frame

zaq32:
* Migrate buffer reader to c++ object, 
* Communication module update.
  * From now on the communication module will first query the frame for its size and then download from the hardware the necessary minimum number of bytes needed.
  * Update CommFrame interface to use the someone else's memory buffer instead of its own fixed size array.
  * Move the frame verification logic out of frame receive procedure to ensure that invalid frame content will block the frame receive path.

---
This change is [Reviewable](https://reviewable.io/reviews/pw-sat2/pwsat2obc/62)

### DIFF
--- a/libs/base/CMakeLists.txt
+++ b/libs/base/CMakeLists.txt
@@ -14,6 +14,10 @@ set(SOURCES
 
 add_library(${NAME} STATIC ${SOURCES})
 
+target_link_libraries(${NAME}
+    gsl
+)
+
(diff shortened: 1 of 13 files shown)


Novakov:
[[Comm] Get exact number of bytes from frame](https://trello.com/c/ZsYdyhFE/90-comm-get-exact-number-of-bytes-from-frame)

Novakov:
Reviewed 13 of 13 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/62)*

PR #65, opened 2017-01-01
File system C++

Novakov:
This change is [Reviewable](https://reviewable.io/reviews/pw-sat2/pwsat2obc/65)


zaq32:
Reviewed 25 of 25 files at r1.
Review status: all files reviewed at latest revision, 13 unresolved discussions.

---

*[libs/fs/extension.cpp, line 58 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/65#-K_Vz4GjxGLx14upZajX:-K_Vz4GjxGLx14upZajY:b-el4c5i) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/3897186932e02be9bf7ac278f75db1204ffa3c2f/libs/fs/extension.cpp#L58)):*
> ```cpp
> }
> 
> File::File(File&& other) : _fs(other._fs), _handle(other._handle), _valid(other._valid)
> ```

mark as noexcept

---

*[libs/fs/extension.cpp, line 64 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/65#-K_Vz5gqCFY9MgT-sr8s:-K_Vz5gqCFY9MgT-sr8t:bvzj871) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/3897186932e02be9bf7ac278f75db1204ffa3c2f/libs/fs/extension.cpp#L64)):*
> ```cpp
> }
> 
> File& File::operator=(File&& other)
> ```

resource leak.

consider this:
```
File foo = File::Open(/* ... */);
foo = File::Open(/* ... */);
```

---

*[libs/fs/extension.cpp, line 64 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/65#-K_VzL4YN0UHT1NxDzYg:-K_VzL4YN0UHT1NxDzYh:b-el4c5i) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/3897186932e02be9bf7ac278f75db1204ffa3c2f/libs/fs/extension.cpp#L64)):*
> ```cpp
> }
> 
> File& File::operator=(File&& other)
> ```

mark as noexcept

---

*[libs/fs/extension.cpp, line 70 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/65#-K_W0npXALRxG0fe09GH:-K_W0npXALRxG0fe09GI:bxdmxan) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/3897186932e02be9bf7ac278f75db1204ffa3c2f/libs/fs/extension.cpp#L70)):*
> ```cpp
>     this->_fs = other._fs;
> 
>     other._handle = 0;
> ```

this suggest that 0 can be used as a marker for invalid handle, with this knowledge you can drop the _valid member from this class.

---

*[libs/fs/yaffs.cpp, line 18 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/65#-K_Vz_WWrTzjRggQYm1g:-K_Vz_WWrTzjRggQYm1h:b-51w6ut) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/3897186932e02be9bf7ac278f75db1204ffa3c2f/libs/fs/yaffs.cpp#L18)):*
> ```cpp
>     else
>     {
>         return (OSResult)yaffs_get_error();
> ```

static_cast

---

*[libs/fs/Include/fs/fs.h, line 64 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/65#-K_W--NgS3w-WIS2ulju:-K_W--NgS3w-WIS2uljv:bnecgxr) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/3897186932e02be9bf7ac278f75db1204ffa3c2f/libs/fs/Include/fs/fs.h#L64)):*
> ```objc
>          * @brief Type that represents directory opening status.
>          */
>         struct DirectoryOpenResult
> ```

the FileOpenResult & DirectoryOpenResult can be merged into single template with one argument (plus typedefs to keep the type names).

---

*[libs/fs/Include/fs/fs.h, line 92 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/65#-K_W00P7u-pqUdHSoouH:-K_W00P7u-pqUdHSoouI:b-34vmyd) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/3897186932e02be9bf7ac278f75db1204ffa3c2f/libs/fs/Include/fs/fs.h#L92)):*
> ```objc
>          * @brief Type that represents file read/write operation status.
>          */
>         struct IOResult
> ```

See above

---

*[libs/fs/Include/fs/fs.h, line 240 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/65#-K_W0ZgbeThBTsRULnb5:-K_W0ZgbeThBTsRULnb6:b-isge0u) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/3897186932e02be9bf7ac278f75db1204ffa3c2f/libs/fs/Include/fs/fs.h#L240)):*
> ```objc
>          * @brief Wrapper over file handle
>          */
>         class File : private NotCopyable
> ```

I'd also add default constructor

---

*[libs/fs/Include/fs/fs.h, line 280 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/65#-K_W1I5gxkjaV5S53BnO:-K_W1I5gxkjaV5S53BnP:b-ko404d) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/3897186932e02be9bf7ac278f75db1204ffa3c2f/libs/fs/Include/fs/fs.h#L280)):*
> ```objc
> 
>             /**
>              * @brief Factory method that opens for read
> ```

for write

---

*[libs/fs/Include/fs/fs.h, line 288 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/65#-K_W1J9WUsm4ahKGNWbv:-K_W1J9WUsm4ahKGNWbw:b28t9op) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/3897186932e02be9bf7ac278f75db1204ffa3c2f/libs/fs/Include/fs/fs.h#L288)):*
> ```objc
>              */
>             static File OpenWrite(
>                 IFileSystem& fs, const char* path, FileOpen mode = FileOpen::Existing, FileAccess access = FileAccess::WriteOnly);
> ```

shouldn't the default mode be CreateAlways?

---

*[libs/fs/Include/fs/fs.h, line 298 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/65#-K_W1T3XreUcTeJxrz2r:-K_W1T3XreUcTeJxrz2s:b-ftig43) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/3897186932e02be9bf7ac278f75db1204ffa3c2f/libs/fs/Include/fs/fs.h#L298)):*
> ```objc
>              * @return Operation result
>              */
>             IOResult Read(gsl::span<uint8_t> buffer);
> ```

it would be more useful to get subspan that contains the actually read data instead of the number of bytes read.

---

*[libs/obc/storage/Include/obc/storage.h, line 38 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/65#-K_W24DyPJqgmcskNbki:-K_W24DyPJqgmcskNbkj:b6x23w5) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/3897186932e02be9bf7ac278f75db1204ffa3c2f/libs/obc/storage/Include/obc/storage.h#L38)):*
> ```objc
>          * @param fs File system interface
>          */
>         OBCStorageHandler(drivers::spi::ISPIInterface& spi, services::fs::YaffsFileSystem& fs);
> ```

why pass implementation instead of interface?

---

*[libs/obc/storage/n25q_flash/n25q.cpp, line 17 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/65#-K_W4bAZvGdLjJRQ6-dC:-K_W4bAZvGdLjJRQ6-dD:b-96knyn) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/3897186932e02be9bf7ac278f75db1204ffa3c2f/libs/obc/storage/n25q_flash/n25q.cpp#L17)):*
> ```cpp
> void N25QStorage::Initialize()
> {
>     if (OS_RESULT_FAILED(this->ExternalFlash.Mount()))
> ```

why external flash is not using IFileSystem interface for volume mounting?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/65)*

zaq32:
Review status: all files reviewed at latest revision, 13 unresolved discussions.

---

*[libs/fs/extension.cpp, line 64 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/65#-K_Vz5gqCFY9MgT-sr8s:-K_W6kPQab1OPCHrlMoJ:b-punevn) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/3897186932e02be9bf7ac278f75db1204ffa3c2f/libs/fs/extension.cpp#L64)):*

resource leak.

consider this:
```
File foo = File::Open(/* ... */);
foo = File::Open(/* ... */);
```
</blockquote></details>

or even more evil:

```
File foo = File::Open(/* ... */);
foo = std::move(foo);
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/65)*

Novakov:
Review status: 12 of 28 files reviewed at latest revision, 13 unresolved discussions.

---

*[libs/fs/extension.cpp, line 58 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/65#-K_Vz4GjxGLx14upZajX:-K__4A4WMZn1IKwXUCaL:b-896fix) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/3897186932e02be9bf7ac278f75db1204ffa3c2f/libs/fs/extension.cpp#L58)):*

mark as noexcept
</blockquote></details>

Done.

---

*[libs/fs/extension.cpp, line 64 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/65#-K_Vz5gqCFY9MgT-sr8s:-K__4LsgOHchifsFpnAA:b-896fix) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/3897186932e02be9bf7ac278f75db1204ffa3c2f/libs/fs/extension.cpp#L64)):*

or even more evil:

```
File foo = File::Open(/* ... */);
foo = std::move(foo);
```
</blockquote></details>

Done.

---

*[libs/fs/extension.cpp, line 64 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/65#-K_VzL4YN0UHT1NxDzYg:-K__4PBkJyFndNIOUg3L:b-896fix) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/3897186932e02be9bf7ac278f75db1204ffa3c2f/libs/fs/extension.cpp#L64)):*

mark as noexcept
</blockquote></details>

Done.

---

*[libs/fs/extension.cpp, line 70 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/65#-K_W0npXALRxG0fe09GH:-K__4SL7Xuhc3DFj1t3n:b-896fix) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/3897186932e02be9bf7ac278f75db1204ffa3c2f/libs/fs/extension.cpp#L70)):*

this suggest that 0 can be used as a marker for invalid handle, with this knowledge you can drop the _valid member from this class.
</blockquote></details>

Done.

---

*[libs/fs/yaffs.cpp, line 18 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/65#-K_Vz_WWrTzjRggQYm1g:-K__4YnR6P4Can54ECfW:b-896fix) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/3897186932e02be9bf7ac278f75db1204ffa3c2f/libs/fs/yaffs.cpp#L18)):*

static_cast
</blockquote></details>

Done.

---

*[libs/fs/Include/fs/fs.h, line 64 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/65#-K_W--NgS3w-WIS2ulju:-K__H_8SVmQW6CYFOnRk:b-896fix) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/3897186932e02be9bf7ac278f75db1204ffa3c2f/libs/fs/Include/fs/fs.h#L64)):*

the FileOpenResult & DirectoryOpenResult can be merged into single template with one argument (plus typedefs to keep the type names).
</blockquote></details>

Done.

---

*[libs/fs/Include/fs/fs.h, line 92 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/65#-K_W00P7u-pqUdHSoouH:-K__H_b67DcIx7ApkTi7:b-896fix) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/3897186932e02be9bf7ac278f75db1204ffa3c2f/libs/fs/Include/fs/fs.h#L92)):*

See above
</blockquote></details>

Done.

---

*[libs/fs/Include/fs/fs.h, line 240 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/65#-K_W0ZgbeThBTsRULnb5:-K__4gGiQwNJF63s4WoW:bj2q7jc) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/3897186932e02be9bf7ac278f75db1204ffa3c2f/libs/fs/Include/fs/fs.h#L240)):*

I'd also add default constructor
</blockquote></details>

File needs reference to `IFileSystem` which cannot be set to some 'empty' value

---

*[libs/fs/Include/fs/fs.h, line 280 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/65#-K_W1I5gxkjaV5S53BnO:-K__86FB5Nc9lM9TSZuY:b-896fix) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/3897186932e02be9bf7ac278f75db1204ffa3c2f/libs/fs/Include/fs/fs.h#L280)):*

for write
</blockquote></details>

Done.

---

*[libs/fs/Include/fs/fs.h, line 288 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/65#-K_W1J9WUsm4ahKGNWbv:-K_W9pNddutY1UM43k9o:b-at4xpv) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/3897186932e02be9bf7ac278f75db1204ffa3c2f/libs/fs/Include/fs/fs.h#L288)):*

shouldn't the default mode be CreateAlways?
</blockquote></details>

To be honest I'm thinking of removing `OpenWrite` - there is no good (and safe) default for open mode 

---

*[libs/fs/Include/fs/fs.h, line 298 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/65#-K_W1T3XreUcTeJxrz2r:-K__Hb5cts1GotIRHGdn:b-896fix) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/3897186932e02be9bf7ac278f75db1204ffa3c2f/libs/fs/Include/fs/fs.h#L298)):*

it would be more useful to get subspan that contains the actually read data instead of the number of bytes read.
</blockquote></details>

Done.

---

*[libs/obc/storage/Include/obc/storage.h, line 38 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/65#-K_W24DyPJqgmcskNbki:-K_WA5fpGwksLT5GlyKH:bnkvxtb) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/3897186932e02be9bf7ac278f75db1204ffa3c2f/libs/obc/storage/Include/obc/storage.h#L38)):*

why pass implementation instead of interface?
</blockquote></details>

`IFileSystem` is independent from yaffs, drivers for storage mechansim are yaffs-only. We can (1) pollute `IFileSystem` with yaffs-specific details, (2) build abstractions to hide yaffs details or (3) pass Yaffs implementation to storage drivers

---

*[libs/obc/storage/n25q_flash/n25q.cpp, line 17 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/65#-K_W4bAZvGdLjJRQ6-dC:-K__81uGf_7W62AEGsYK:b-896fix) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/3897186932e02be9bf7ac278f75db1204ffa3c2f/libs/obc/storage/n25q_flash/n25q.cpp#L17)):*

why external flash is not using IFileSystem interface for volume mounting?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/65)*

zaq32:
Reviewed 16 of 16 files at r2.
Review status: all files reviewed at latest revision, 8 unresolved discussions.

---

*[libs/fs/extension.cpp, line 64 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/65#-K_Vz5gqCFY9MgT-sr8s:-K_eodnVQ5dO07XzPygy:b-y5ml3m) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/3897186932e02be9bf7ac278f75db1204ffa3c2f/libs/fs/extension.cpp#L64)):*

Done.
</blockquote></details>


I am afraid not current implementation is actually worse than the original one as contains the resource leak plus it has unintended side effect of leaving the resource File object in invalid state. 

Take the first example:

```
File foo = File::Open(/* ... */);   // (1)
foo = File::Open(/* ... */);          // (2)
```

at line 1 we have valid opened file which is indicated by the _handle member set to to some value lets use symbol A for it. 
so at line 1 we have
```
foo:
_handle : A
_valid : true
```
now we at line 2 we are opening the new file. The File::Open is returning another File object which is move assigned to the foo variable. At the point of assignment in line two we have:
```
foo: 
_handle: A
_valid: true

<temporary>:
_handle: B
_valid: true
```

Now we enter the move assignment operator where this points to ```foo``` and other is bound to ```<temporary>```

Lets see what happens inside the operator:

```
this->_handle = other._handle;        // foo->_handle <- B   the handle with value A is leaked here (!!!)
this->_valid = other._valid;          // foo->_valid <- true   
this->_fs = other._fs;                // this line is subject for another discussion
// here is where it gets very interesting
other._handle = this->_handle;       // <temporary>->_handle <- foo->_handle which is at this point B as it was set two line above
other._valid = this->_valid;         // <temporary>->_valid <- foo->_valid 
other._fs = this->_fs;
```
so after the assignment and before end of this line we have following situation:
```
foo: 
_handle: B
_valid: true

<temporary>:
_handle: B
_valid: true
```
and what happens now is the destructor for the ```<temporary>``` gets called which closes the resource B(!!!) as a result we end up with leaked file handle A and foo object that uses already destroyed file handle B.

---

*[libs/fs/extension.cpp, line 70 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/65#-K_W0npXALRxG0fe09GH:-K_et76-yhLswHbfs-VM:b-abbzcs) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/3897186932e02be9bf7ac278f75db1204ffa3c2f/libs/fs/extension.cpp#L70)):*

Done.
</blockquote></details>

This is not necessarily what I have suggested. Since 0 has been used to indicate that handle is not valid the value of the _valid member can be inferred by checking if the _handle member is set to 0 as a result the _valid member can be completely removed.

---

*[libs/fs/extension.cpp, line 58 at r2](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/65#-K_evWRykAJSxFUyN5Bx:-K_evWRykAJSxFUyN5By:bqwpluy) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/4b57803b14bc56b9b6ffa17a949360fcd3082228/libs/fs/extension.cpp#L58)):*
> ```cpp
> }
> 
> File::File(File&& other) noexcept : _fs(other._fs), _handle(other._handle), _valid(other._valid)
> ```

see my comment regarding move assignment operator. This constructor is now suffering for the same issue.

---

*[libs/fs/Include/fs/fs.h, line 240 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/65#-K_W0ZgbeThBTsRULnb5:-K_etafcgRY4trFS_fOc:b5jshh6) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/3897186932e02be9bf7ac278f75db1204ffa3c2f/libs/fs/Include/fs/fs.h#L240)):*

File needs reference to `IFileSystem` which cannot be set to some 'empty' value
</blockquote></details>

Since it is a reference the File type should not be moveable as well because it is not possible to redirect once initialized reference.

Consider this:

```
int a = 1;
int& b = a;
int* c = &a;
int d = 4;
```

now ```b = d;``` is semantically equal to the ```*c = d;```
Also the redirection possible with pointers like this: ```c = &d;``` is not possible to replicate with references. 

The reason why the _fs member assignment works at all in the move constructor is caused only by the fact that in current system _fs in both source & the target variables will always point to the same service. However you will quickly start seeing unintended side effects once you start using several file system services. 

---

*[libs/fs/Include/fs/fs.h, line 288 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/65#-K_W1J9WUsm4ahKGNWbv:-K_eo5jQaKW7lBLRRbLu:barp6or) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/3897186932e02be9bf7ac278f75db1204ffa3c2f/libs/fs/Include/fs/fs.h#L288)):*

To be honest I'm thinking of removing `OpenWrite` - there is no good (and safe) default for open mode 
</blockquote></details>

why not just drop the default value from  opening mode argument?

---

*[libs/obc/storage/Include/obc/storage.h, line 38 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/65#-K_W24DyPJqgmcskNbki:-K_evvKQD9VrS_0QusDu:b-nnr5vr) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/3897186932e02be9bf7ac278f75db1204ffa3c2f/libs/obc/storage/Include/obc/storage.h#L38)):*

`IFileSystem` is independent from yaffs, drivers for storage mechansim are yaffs-only. We can (1) pollute `IFileSystem` with yaffs-specific details, (2) build abstractions to hide yaffs details or (3) pass Yaffs implementation to storage drivers
</blockquote></details>

what about (4) create dedicated yaffs specific interface directed at the file system providers? My point here is that I would like to see some automatic tests for the cases when storage device could not be mounted.

---

*[libs/obc/storage/n25q_flash/n25q.cpp, line 19 at r2](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/65#-K_em6IoiYffgqc2ay9s:-K_em6IoiYffgqc2ay9t:bet3jvm) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/4b57803b14bc56b9b6ffa17a949360fcd3082228/libs/obc/storage/n25q_flash/n25q.cpp#L19)):*
> ```cpp
>     if (OS_RESULT_FAILED(this->ExternalFlash.Mount(this->_fs)))
>     {
>         return;
> ```

I'd recommend returning status value that would indicate that flash is not available. so we can take proper action at the top level and perhaps adjust list of the available services.

---

*[libs/obc/storage/stk_flash/stk.cpp, line 42 at r2](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/65#-K_el_UhmIlQow3tH-H3:-K_el_UhmIlQow3tH-H4:bet3jvm) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/4b57803b14bc56b9b6ffa17a949360fcd3082228/libs/obc/storage/stk_flash/stk.cpp#L42)):*
> ```cpp
>     if (OS_RESULT_FAILED(this->_fs.AddDeviceAndMount(&rootDevice)))
>     {
>         return;
> ```

I'd recommend returning status value that would indicate that flash is not available. so we can take proper action at the top level and perhaps adjust list of the available services.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/65)*

Novakov:
Review status: 15 of 28 files reviewed at latest revision, 8 unresolved discussions.

---

*[libs/fs/extension.cpp, line 64 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/65#-K_Vz5gqCFY9MgT-sr8s:-K_hvbcJQOCErQBezGWZ:bsd61vr) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/3897186932e02be9bf7ac278f75db1204ffa3c2f/libs/fs/extension.cpp#L64)):*


I am afraid not current implementation is actually worse than the original one as contains the resource leak plus it has unintended side effect of leaving the resource File object in invalid state. 

Take the first example:

```
File foo = File::Open(/* ... */);   // (1)
foo = File::Open(/* ... */);          // (2)
```

at line 1 we have valid opened file which is indicated by the _handle member set to to some value lets use symbol A for it. 
so at line 1 we have
```
foo:
_handle : A
_valid : true
```
now we at line 2 we are opening the new file. The File::Open is returning another File object which is move assigned to the foo variable. At the point of assignment in line two we have:
```
foo: 
_handle: A
_valid: true

<temporary>:
_handle: B
_valid: true
```

Now we enter the move assignment operator where this points to ```foo``` and other is bound to ```<temporary>```

Lets see what happens inside the operator:

```
this->_handle = other._handle;        // foo->_handle <- B   the handle with value A is leaked here (!!!)
this->_valid = other._valid;          // foo->_valid <- true   
this->_fs = other._fs;                // this line is subject for another discussion
// here is where it gets very interesting
other._handle = this->_handle;       // <temporary>->_handle <- foo->_handle which is at this point B as it was set two line above
other._valid = this->_valid;         // <temporary>->_valid <- foo->_valid 
other._fs = this->_fs;
```
so after the assignment and before end of this line we have following situation:
```
foo: 
_handle: B
_valid: true

<temporary>:
_handle: B
_valid: true
```
and what happens now is the destructor for the ```<temporary>``` gets called which closes the resource B(!!!) as a result we end up with leaked file handle A and foo object that uses already destroyed file handle B.
</blockquote></details>

I really wrote that? I really need more sleep/coffee.

---

*[libs/fs/extension.cpp, line 70 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/65#-K_W0npXALRxG0fe09GH:-K_ho0R_0WqaCdtIzBhj:b8vl2yp) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/3897186932e02be9bf7ac278f75db1204ffa3c2f/libs/fs/extension.cpp#L70)):*

This is not necessarily what I have suggested. Since 0 has been used to indicate that handle is not valid the value of the _valid member can be inferred by checking if the _handle member is set to 0 as a result the _valid member can be completely removed.
</blockquote></details>

I never intended to use `_handle` equal of 0 as error marker - that's why I'm using two values. Assigning 0 to `_handle` was by mistake

---

*[libs/fs/Include/fs/fs.h, line 240 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/65#-K_W0ZgbeThBTsRULnb5:-K_hvvWn_jtjrArbaf9p:b-aaynk9) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/3897186932e02be9bf7ac278f75db1204ffa3c2f/libs/fs/Include/fs/fs.h#L240)):*

Since it is a reference the File type should not be moveable as well because it is not possible to redirect once initialized reference.

Consider this:

```
int a = 1;
int& b = a;
int* c = &a;
int d = 4;
```

now ```b = d;``` is semantically equal to the ```*c = d;```
Also the redirection possible with pointers like this: ```c = &d;``` is not possible to replicate with references. 

The reason why the _fs member assignment works at all in the move constructor is caused only by the fact that in current system _fs in both source & the target variables will always point to the same service. However you will quickly start seeing unintended side effects once you start using several file system services. 


</blockquote></details>

So no move operations => no factory methods

---

*[libs/obc/storage/Include/obc/storage.h, line 38 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/65#-K_W24DyPJqgmcskNbki:-K_i6PSpyRN0kGe_EApt:bb6un1j) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/3897186932e02be9bf7ac278f75db1204ffa3c2f/libs/obc/storage/Include/obc/storage.h#L38)):*

what about (4) create dedicated yaffs specific interface directed at the file system providers? My point here is that I would like to see some automatic tests for the cases when storage device could not be mounted.
</blockquote></details>

Done

---

*[libs/obc/storage/n25q_flash/n25q.cpp, line 19 at r2](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/65#-K_em6IoiYffgqc2ay9s:-K_i2-4BKPZcPyHKF7AD:b-896fix) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/4b57803b14bc56b9b6ffa17a949360fcd3082228/libs/obc/storage/n25q_flash/n25q.cpp#L19)):*

I'd recommend returning status value that would indicate that flash is not available. so we can take proper action at the top level and perhaps adjust list of the available services.
</blockquote></details>

Done.

---

*[libs/obc/storage/stk_flash/stk.cpp, line 42 at r2](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/65#-K_el_UhmIlQow3tH-H3:-K_i20aYKtlM2xthTXJJ:b-896fix) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/4b57803b14bc56b9b6ffa17a949360fcd3082228/libs/obc/storage/stk_flash/stk.cpp#L42)):*

I'd recommend returning status value that would indicate that flash is not available. so we can take proper action at the top level and perhaps adjust list of the available services.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/65)*

zaq32:
Reviewed 13 of 13 files at r3.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[libs/fs/extension.cpp, line 64 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/65#-K_Vz5gqCFY9MgT-sr8s:-K_kOKyLCf69GkE5iFQL:b-obnuv3) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/3897186932e02be9bf7ac278f75db1204ffa3c2f/libs/fs/extension.cpp#L64)):*

I really wrote that? I really need more sleep/coffee.
</blockquote></details>

That was drastic...
but will work

---

*[libs/fs/Include/fs/fs.h, line 240 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/65#-K_W0ZgbeThBTsRULnb5:-K_kOf6enUaI1BRb9dc9:b-gzqpny) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/3897186932e02be9bf7ac278f75db1204ffa3c2f/libs/fs/Include/fs/fs.h#L240)):*

So no move operations => no factory methods
</blockquote></details>

You may also internally use pointers and keep the reference in the interface

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/65)*

PR #77, opened 2017-01-26
Changed meaning for I2C adress to 7-bit value

ggajoch:
After physical tests it turned out that adresses in ICDs are 7-bit value without read/write marker padding.
Therefore, meaning of I2CAdress in I2C interface was changes to reflect 7-bit adress standard.

---
This change is [Reviewable](https://reviewable.io/reviews/pw-sat2/pwsat2obc/77)

### DIFF
--- a/integration_tests/devices/antenna.py
+++ b/integration_tests/devices/antenna.py
@@ -2,8 +2,8 @@ import logging
 import i2cMock
 from utils import *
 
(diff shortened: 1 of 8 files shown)


zaq32:
Reviewed 8 of 8 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/77)*

PR #107, opened 2017-03-22
[comm] Prevent beacon update from erasing frames from transmitter queue

zaq32:
Update beacon setting procedure to ensure that setting beacon will not cause
transmitter to remove any frames from the outgoing frame buffer before they
are send. This is done by first sending beacon as a regular frame and checking
the number of transmitter''s output buffer free slot count.

---
This change is [Reviewable](https://reviewable.io/reviews/pw-sat2/pwsat2obc/107)

### DIFF
--- a/libs/drivers/comm/Include/comm/CommDriver.hpp
+++ b/libs/drivers/comm/Include/comm/CommDriver.hpp
@@ -8,6 +8,7 @@
 #include "base/os.h"
 #include "comm.hpp"
 #include "error_counter/error_counter.hpp"
+#include "gsl/span"
 #include "i2c/forward.h"
 
 COMM_BEGIN
@@ -123,6 +124,9 @@ class CommObject final : public ITransmitFrame, public IBeaconController
     /**
      * @brief This procedure sets the beacon frame for the passed comm object.
      *
+     * This procedure will ensure that setting beacon will not erase any frames from
+     * transmitter's output buffer by first sending beacon as a regular frame and setting
+     * the beacon itself only if there are no frames that are waiting to te send by the transmitter.
      * @param[in] beacon Reference to object describing new beacon.
      * See the definition of the CommBeacon for details.
      * @return Operation status, true in case of success, false otherwise.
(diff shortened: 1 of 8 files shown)


zaq32:
[[comm] Beacon update process should not remove any frames from transmitter queue](https://trello.com/c/k59qCrO3/146-comm-beacon-update-process-should-not-remove-any-frames-from-transmitter-queue)

Novakov:
Reviewed 4 of 4 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[libs/drivers/comm/Include/comm/CommDriver.hpp, line 278 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/107#-KgA6C33YVxnIP7shhlC:-KgA6C33YVxnIP7shhlD:b-2y8tu) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/56fb5a683cf2747528890663c7932988fb0055a1/libs/drivers/comm/Include/comm/CommDriver.hpp#L278)):*
> ```cpp
>      * @return Operation status, true in case of success, false otherwise.
>      */
>     bool ScheduleFrameTransmission(gsl::span<const std::uint8_t> frame, std::uint8_t& remainingSlots);
> ```

How about using `Option<T>` as return type?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/107)*

zaq32:
Review status: 1 of 8 files reviewed at latest revision, 1 unresolved discussion.

---

*[libs/drivers/comm/Include/comm/CommDriver.hpp, line 278 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/107#-KgA6C33YVxnIP7shhlC:-KgBVIPXBKVCOFkrGbfJ:b-25gr9n) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/56fb5a683cf2747528890663c7932988fb0055a1/libs/drivers/comm/Include/comm/CommDriver.hpp#L278)):*

How about using `Option<T>` as return type?
</blockquote></details>

In here it doesn't make sense as this method is pure command implementation. Did you have the "SetBeacon" method in mind?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/107)*

Novakov:
Reviewed 7 of 7 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/107)*

Novakov:
Reviewed 2 of 2 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/107)*

PR #125, opened 2017-04-08
You've been calculating your stack size WRONG whole time - click here to see why!

Novakov:
`xTaskCreate` expects parameter with size of the stack for new task in **words**. However size of the **word** is port-dependent and in case of Cortex-M3 its 4 bytes not 2 as we assumed at the begining. 

This change hides calculation of size in words by doing the division just before calling `xTaskCreate` and using size of `StackType_t`.

---
This change is [Reviewable](https://reviewable.io/reviews/pw-sat2/pwsat2obc/125)

### DIFF
--- a/libs/base/Include/base/os.h
+++ b/libs/base/Include/base/os.h
@@ -206,7 +206,7 @@ class System final : public PureStatic
      *
      * @param[in] entryPoint Pointer to task procedure.
      * @param[in] taskName Name of the new task.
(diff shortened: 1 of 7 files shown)


zaq32:
Reviewed 7 of 7 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/125)*

PR #127, opened 2017-04-08
[WIP] Time synchronization update

zaq32:
* Remove time service's serialization ability & replace it with addition to the satellite persistent state.
* Update time synchronization component to make use of persistent state instead of its own in-memory state. As a result time synchronization has been simplified a bit due to it becoming stateless. In addition as a result of this change from now on time lost due to obc being powerred of is now at least partially corrected (50% due to the  mcu time being stopped during this period).
* Introduced time correction configuration. This change provides possibility to adjust time synchronization algorithm in order to make it more resilient to permanent/increasing time measuring errors that may be detected after sattelite launch. It is also possible to disable one of the time sources if it starts to generate too many errors.
* Extend, update persistent state & time synchronization documentation.

Awaits for integration of persistent state object into mainline.

---
This change is [Reviewable](https://reviewable.io/reviews/pw-sat2/pwsat2obc/127)


zaq32:
[[Time] Remove auto saving from TimeProvider](https://trello.com/c/gBCiUIJ3/133-time-remove-auto-saving-from-timeprovider)

zaq32:
[[Time] Time correction should use parametrized weight](https://trello.com/c/JKCVUPpD/151-time-time-correction-should-use-parametrized-weight)

zaq32:
[[Time] Use FRAM to store value of last seen RTC value](https://trello.com/c/7TyINcDS/130-time-use-fram-to-store-value-of-last-seen-rtc-value)

PR #197, opened 2017-06-29
Payload driver

tom4r:
Created Payload Driver.

---
This change is [Reviewable](https://reviewable.io/reviews/pw-sat2/pwsat2obc/197)


Novakov:
Reviewed 25 of 53 files at r1.
Review status: 25 of 53 files reviewed at latest revision, 2 unresolved discussions.

---

*[unit_tests/base/Include/mock/InterruptPinDriverMock.hpp, line 11 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197#-Knnj4GL7OSzlzF4SBQr:-Knnj4GL7OSzlzF4SBQs:brhcwa) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/3095f42da92104d1b02b32667aa3a55521326026/unit_tests/base/Include/mock/InterruptPinDriverMock.hpp#L11)):*
> ```cpp
> using testing::Invoke;
> 
> class InterruptPinDriverMock : public drivers::gpio::IInterruptPinDriver
> ```

Dont' define mock constructors inline - it makes compilation/linking longer

---

*[unit_tests/base/Include/mock/PayloadHardwareDriverMock.hpp, line 12 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197#-KnnjMg0wwTW_foFKVew:-KnnjMg0wwTW_foFKVex:b-v0gx9i) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/3095f42da92104d1b02b32667aa3a55521326026/unit_tests/base/Include/mock/PayloadHardwareDriverMock.hpp#L12)):*
> ```cpp
> {
>   public:
>     MOCK_CONST_METHOD0(IsBusy, bool());
> ```

Don't define mock constructors inline

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197)*

ggajoch:
Review status: 25 of 53 files reviewed at latest revision, 4 unresolved discussions.

---

*[libs/drivers/payload/Include/payload/devices.h, line 40 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197#-KnoKiVtr3DiRcy0mNf7:-KnoKiVtr3DiRcy0mNf8:bwdeho) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/3095f42da92104d1b02b32667aa3a55521326026/libs/drivers/payload/Include/payload/devices.h#L40)):*
> ```objc
>              * @return Result status.
>              */
>             virtual OSResult GetWhoami(PayloadTelemetry::Status& output) override;
> ```

shouldn't there be also a method to check value of WhoAmI register?
When PLD will be turned on OBC should check this register for correctness, in addition to writing it to experiment data.
However, I am not sure if we should abort the experiment in case PLD will respond (ACK on I2C) but returned incorrect WhoAmI data.

---

*[libs/drivers/payload/Include/payload/commands/base_code.hpp, line 46 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197#-KnoJy6RN0td1dWu-RYV:-KnoJy6RN0td1dWu-RYW:b-3muvi6) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/3095f42da92104d1b02b32667aa3a55521326026/libs/drivers/payload/Include/payload/commands/base_code.hpp#L46)):*
> ```cpp
> 
>     // Validation should not cancel save.
>     result = Validate();
> ```

validation result should be ignored (as it is right now) and performed only on ground.
Then, is the validating code needed here?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197)*

zaq32:
Reviewed 31 of 53 files at r1.
Review status: 31 of 53 files reviewed at latest revision, 10 unresolved discussions.

---

*[libs/drivers/gpio/Include/gpio/InterruptPinDriver.h, line 60 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197#-Knsb43QcIhdqU678TWS:-Knsb43QcIhdqU678TWT:b-px3c22) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/3095f42da92104d1b02b32667aa3a55521326026/libs/drivers/gpio/Include/gpio/InterruptPinDriver.h#L60)):*
> ```objc
>              * @brief Returns mask for setting and clearing interrupt registers. Calculated from interrupt pin number.
>              */
>             inline uint32_t IRQMask()
> ```

define outside class definition

---

*[libs/drivers/payload/Include/payload/payload.h, line 94 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197#-KnsZ5VLGmoP7pOinZrd:-KnsZ5VMJcZGS4y9ht7k:betxbq4) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/3095f42da92104d1b02b32667aa3a55521326026/libs/drivers/payload/Include/payload/payload.h#L94)):*
> ```objc
> 
>           private:
>             static constexpr uint32_t InterruptPriority = 6;
> ```

values like this should not be hardcoded inside the driver.

---

*[libs/drivers/payload/Include/payload/telemetry.h, line 88 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197#-KnsZHtQSxPfYZ-Q9dhM:-KnsZHtQSxPfYZ-Q9dhN:bep413g) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/3095f42da92104d1b02b32667aa3a55521326026/libs/drivers/payload/Include/payload/telemetry.h#L88)):*
> ```objc
>                 std::uint32_t temperature;
>                 std::array<std::uint32_t, 3> vth;
>             } __attribute__((packed));
> ```

what is the purpose of this attribute?

---

*[libs/drivers/payload/Include/payload/commands/base.h, line 25 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197#-Kns_19gvVG3z2RGDXBX:-Kns_19gvVG3z2RGDXBY:bnsvcl4) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/3095f42da92104d1b02b32667aa3a55521326026/libs/drivers/payload/Include/payload/commands/base.h#L25)):*
> ```objc
>                  * @param[in] driver A reference to hardware driver for payload.
>                  */
>                 PayloadCommand(IPayloadDriver& driver) : _driver(driver)
> ```

define constructor outside template definition

---

*[libs/drivers/payload/Include/payload/commands/base.h, line 74 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197#-Kns_BazpAPTOlZDeu4M:-Kns_BazpAPTOlZDeu4N:b-g9wfeq) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/3095f42da92104d1b02b32667aa3a55521326026/libs/drivers/payload/Include/payload/commands/base.h#L74)):*
> ```objc
>                  * @brief The hardware driver.
>                  */
>                 IPayloadDriver& _driver;
> ```

do not use public member fields

---

*[src/obc.cpp, line 103 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197#-KnsaAW8K7MIEuaNGOjp:-KnsaAW8K7MIEuaNGOjq:b45y73c) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/f1bc8007ba906abbcf7f17fea8f3587196b277f4/src/obc.cpp#L103)):*
> ```cpp
>     {
>         LOGF(LOG_LEVEL_FATAL, "Storage initialization failed %d", num(result));
>         return result;
> ```

why this has been removed?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197)*

tom4r:
Review status: 21 of 56 files reviewed at latest revision, 10 unresolved discussions.

---

*[libs/drivers/gpio/Include/gpio/InterruptPinDriver.h, line 60 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197#-Knsb43QcIhdqU678TWS:-KnzmNfKM_r5-PzTMc1I:b-886gsx) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/3095f42da92104d1b02b32667aa3a55521326026/libs/drivers/gpio/Include/gpio/InterruptPinDriver.h#L60)):*

define outside class definition
</blockquote></details>

Fixed

---

*[libs/drivers/payload/Include/payload/devices.h, line 40 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197#-KnoKiVtr3DiRcy0mNf7:-KnzmbUlhnTHBhbUHVnt:b4wtld3) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/3095f42da92104d1b02b32667aa3a55521326026/libs/drivers/payload/Include/payload/devices.h#L40)):*
<details><summary><i>Previously, ggajoch (Grzegorz Gajoch) wrote…</i></summary><blockquote>

shouldn't there be also a method to check value of WhoAmI register?
When PLD will be turned on OBC should check this register for correctness, in addition to writing it to experiment data.
However, I am not sure if we should abort the experiment in case PLD will respond (ACK on I2C) but returned incorrect WhoAmI data.

</blockquote></details>

Added ValidateWhoami method for checking if value returned from payload is correct.

---

*[libs/drivers/payload/Include/payload/payload.h, line 94 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197#-KnsZ5VLGmoP7pOinZrd:-KnzmmaH94txx8iuOuTd:b-gtbaf1) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/3095f42da92104d1b02b32667aa3a55521326026/libs/drivers/payload/Include/payload/payload.h#L94)):*

values like this should not be hardcoded inside the driver.
</blockquote></details>

Moved to io_map like for example in UART driver.

---

*[libs/drivers/payload/Include/payload/telemetry.h, line 88 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197#-KnsZHtQSxPfYZ-Q9dhM:-KnzmtbidT6YjgQu2joH:bej44ne) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/3095f42da92104d1b02b32667aa3a55521326026/libs/drivers/payload/Include/payload/telemetry.h#L88)):*

what is the purpose of this attribute?
</blockquote></details>

This is the GCC-specific attribute that removes byte(s) of padding between structure fields. Here used for removing padding between status and temperature fields. 

---

*[libs/drivers/payload/Include/payload/commands/base.h, line 25 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197#-Kns_19gvVG3z2RGDXBX:-KnznRR1F_xQLCCQuIsc:bt2vr1b) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/3095f42da92104d1b02b32667aa3a55521326026/libs/drivers/payload/Include/payload/commands/base.h#L25)):*

define constructor outside template definition
</blockquote></details>

Fixed.

---

*[libs/drivers/payload/Include/payload/commands/base.h, line 74 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197#-Kns_BazpAPTOlZDeu4M:-KnznU6El-Gxf-5oCNlw:b-vlgd8t) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/3095f42da92104d1b02b32667aa3a55521326026/libs/drivers/payload/Include/payload/commands/base.h#L74)):*

do not use public member fields
</blockquote></details>

This field is protected.

---

*[libs/drivers/payload/Include/payload/commands/base_code.hpp, line 46 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197#-KnoJy6RN0td1dWu-RYV:-KnznlTrn0bVpN-nw1wg:b-ds1aud) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/3095f42da92104d1b02b32667aa3a55521326026/libs/drivers/payload/Include/payload/commands/base_code.hpp#L46)):*
<details><summary><i>Previously, ggajoch (Grzegorz Gajoch) wrote…</i></summary><blockquote>

validation result should be ignored (as it is right now) and performed only on ground.
Then, is the validating code needed here?
</blockquote></details>

Removed all validation code in commands.

---

*[src/obc.cpp, line 103 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197#-KnsaAW8K7MIEuaNGOjp:-Knzo0SzjL7T3t2F3i0U:b3p45ks) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/f1bc8007ba906abbcf7f17fea8f3587196b277f4/src/obc.cpp#L103)):*

why this has been removed?
</blockquote></details>

It was removed after discussion with Maciek, because currently it causes hardfault several lines of code later in scenario when Flash completely doesn't work or is not connected (like in my first setup without payload board), because it skips the rest of required initialization.
With return removed, at least OBC starts.

---

*[unit_tests/base/Include/mock/InterruptPinDriverMock.hpp, line 11 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197#-Knnj4GL7OSzlzF4SBQr:-KnzoX4gg6onh5-rf0AN:bt2vr1b) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/3095f42da92104d1b02b32667aa3a55521326026/unit_tests/base/Include/mock/InterruptPinDriverMock.hpp#L11)):*

Dont' define mock constructors inline - it makes compilation/linking longer
</blockquote></details>

Fixed.

---

*[unit_tests/base/Include/mock/PayloadHardwareDriverMock.hpp, line 12 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197#-KnnjMg0wwTW_foFKVew:-Knzo_iBF-eYOKQmnZfE:bt2vr1b) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/3095f42da92104d1b02b32667aa3a55521326026/unit_tests/base/Include/mock/PayloadHardwareDriverMock.hpp#L12)):*

Don't define mock constructors inline
</blockquote></details>

Fixed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197)*

Novakov:
Reviewed 1 of 30 files at r2.
Review status: 22 of 56 files reviewed at latest revision, 9 unresolved discussions.

---

*[libs/drivers/payload/Include/payload/commands/Housekeeping.h, line 35 at r2](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197#-Ko1sZSuY1_BTEWh4Reg:-Ko1sZSuY1_BTEWh4Reh:bins4ih) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/630638c2c28090e203637aa8a3755e33ad2361e9/libs/drivers/payload/Include/payload/commands/Housekeeping.h#L35)):*
> ```objc
>                     std::array<uint8_t, sizeof(PayloadTelemetry::Housekeeping)> buffer;
>                     static_assert(sizeof(data) == sizeof(buffer), "Incorrect size buffered Telemetry");
>                 } _telemetry;
> ```

It's bad idea to relay on struct members order when reading response from any device. Normally we use following procedure:
1) Read response bytes into buffer
2) Use `Reader` to extract values from buffer into dedicated response structure

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197)*

tom4r:
Review status: 22 of 56 files reviewed at latest revision, 9 unresolved discussions.

---

*[libs/drivers/payload/Include/payload/commands/Housekeeping.h, line 35 at r2](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197#-Ko1sZSuY1_BTEWh4Reg:-Ko2vWLgpPJ17IoQUjJN:b-99hrsm) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/630638c2c28090e203637aa8a3755e33ad2361e9/libs/drivers/payload/Include/payload/commands/Housekeeping.h#L35)):*

It's bad idea to relay on struct members order when reading response from any device. Normally we use following procedure:
1) Read response bytes into buffer
2) Use `Reader` to extract values from buffer into dedicated response structure
</blockquote></details>

OK. I "demagicified" the "I2C data serialization" code. :)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197)*

Novakov:
Reviewed 4 of 53 files at r1, 15 of 30 files at r2, 15 of 15 files at r3.
Review status: all files reviewed at latest revision, 19 unresolved discussions, some commit checks failed.

---

*[integration_tests/devices/payload.py, line 22 at r3](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197#-Ko8bQ5N99jA3-J62fzU:-Ko8bQ5N99jA3-J62fzV:bjj545a) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/e770aadb7869b677dba0c5ee4c6c37964b1fe36b/integration_tests/devices/payload.py#L22)):*
> ```Python
>         self.RadFET = [0xFF, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF]
>         self.default_processing_time = 1
>         self.long_processing_time = 10
> ```

It will be better to provide callback methods returning processing time (or no time at all) 

---

*[integration_tests/devices/payload.py, line 195 at r3](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197#-Ko8aY1w2I1R35GPG_os:-Ko8aY1w2I1R35GPG_ot:byv2dwg) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/e770aadb7869b677dba0c5ee4c6c37964b1fe36b/integration_tests/devices/payload.py#L195)):*
> ```Python
>     @i2cMock.command([41])
>     def read_radfet(self):
>         self.log.debug("Read RadFET")
> ```

By default all calls to I2C Device methods are logged - no need to log them second time

---

*[libs/drivers/payload/commands/SunS.cpp, line 35 at r3](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197#-Ko8nuzvmRG9IJuRKay8:-Ko8nuzvmRG9IJuRKay9:b-8pc2qc) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/e770aadb7869b677dba0c5ee4c6c37964b1fe36b/libs/drivers/payload/commands/SunS.cpp#L35)):*
> ```cpp
>     {
>         LOGF(LOG_LEVEL_DEBUG, "SunS Ref voltage %u: %u", i + 1, output.voltages[i]);
>     }
> ```

I don't think it is necessary to log every received value

---

*[libs/drivers/payload/commands/Whoami.cpp, line 17 at r3](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197#-Ko8nKG8frN2GANxt6Ok:-Ko8nKG8frN2GANxt6Ol:b4izlh0) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/e770aadb7869b677dba0c5ee4c6c37964b1fe36b/libs/drivers/payload/commands/Whoami.cpp#L17)):*
> ```cpp
>     if (_driver.IsBusy())
>     {
>         LOG(LOG_LEVEL_WARNING, "Payload busy. Ignoring command");
> ```

We tend to use `[module] Message` convention for log message - just a suggestion

---

*[libs/drivers/payload/Include/payload/payload.h, line 97 at r3](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197#-Ko8pTBNgJ_kHFZ7PuQL:-Ko8pTBOTsOkjXmDotsC:b-il86eb) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/e770aadb7869b677dba0c5ee4c6c37964b1fe36b/libs/drivers/payload/Include/payload/payload.h#L97)):*
> ```objc
>              * @brief DefaultTimeout Default timeout for Payload operations. 30 minutes.
>              */
>             static constexpr std::chrono::milliseconds DefaultTimeout = std::chrono::milliseconds(1800000);
> ```

30 minutes is quite a long time...

---

*[libs/drivers/payload/Include/payload/telemetry.h, line 87 at r3](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197#-Ko8p5re-ruo2MBw0Lin:-Ko8p5re-ruo2MBw0Lio:b9cnjaf) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/e770aadb7869b677dba0c5ee4c6c37964b1fe36b/libs/drivers/payload/Include/payload/telemetry.h#L87)):*
> ```objc
>                 std::uint16_t obc_3v3d;
>                 static constexpr uint8_t DeviceDataAddress = 37;
>                 static constexpr uint8_t DeviceDataLength = 4;
> ```

You can streamline form of this code a bit by defining base struct taking address and length as template parameters and exposing them as constexprs

---

*[libs/drivers/payload/Include/payload/telemetry.h, line 103 at r3](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197#-Ko8oyAF-gmJ0veMjUlz:-Ko8oyAF-gmJ0veMjUm-:btw6er8) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/e770aadb7869b677dba0c5ee4c6c37964b1fe36b/libs/drivers/payload/Include/payload/telemetry.h#L103)):*
> ```objc
>         };
> 
>         static_assert(std::is_pod<PayloadTelemetry>::value, "POD");
> ```

As you are no longer relaying on memory layout you can remove this assert

---

*[libs/drivers/payload/Include/payload/commands/base.h, line 75 at r3](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197#-Ko8oWctmg_LkWWCvqK4:-Ko8oWctmg_LkWWCvqK5:be3zpew) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/e770aadb7869b677dba0c5ee4c6c37964b1fe36b/libs/drivers/payload/Include/payload/commands/base.h#L75)):*
> ```objc
>         }
>     }
> }
> ```

I think it is possible to remove almost all virtual calls by using constexprs inside command classes (for data address and buffer). Most of the code in commands is the same

---

*[libs/drivers/payload/Include/payload/commands/base_code.hpp, line 86 at r3](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197#-Ko8mS7ClTNB7qSZODDH:-Ko8mS7ClTNB7qSZODDI:b-mjn8gp) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/e770aadb7869b677dba0c5ee4c6c37964b1fe36b/libs/drivers/payload/Include/payload/commands/base_code.hpp#L86)):*
> ```cpp
> 
>     return OSResult::Success;
> }
> ```

why splitting into two files?

---

*[platforms/FlightModel/Include/io_map.h, line 32 at r3](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197#-Ko8ifV6Q57x1D5MU5Ss:-Ko8ifV6Q57x1D5MU5St:b-2hd1ca) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/e770aadb7869b677dba0c5ee4c6c37964b1fe36b/platforms/FlightModel/Include/io_map.h#L32)):*
> ```objc
>     {
>         static constexpr std::uint8_t InterruptPriority = 6;
>         using PayloadInterrupt = PinLocation<gpioPortE, 0>;
> ```

During last tests we found out that different pin is used for payload interrupt - verify with @ggajoch and @begreeen 

---

*[src/commands/payload.cpp, line 233 at r3](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197#-Ko8jXIRlkkOqhDRFpsu:-Ko8jXIRlkkOqhDRFpsv:bancfph) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/e770aadb7869b677dba0c5ee4c6c37964b1fe36b/src/commands/payload.cpp#L233)):*
> ```cpp
> 
>     cmd(driver, argc - 1, &argv[1]);
> }
> ```

In all commands: provide labels for each value in output. It's much easier to understand (and determine if it's correct)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197)*

tom4r:
Review status: 28 of 56 files reviewed at latest revision, 19 unresolved discussions, some commit checks failed.

---

*[integration_tests/devices/payload.py, line 22 at r3](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197#-Ko8bQ5N99jA3-J62fzU:-KoLi64fqgnZ88Uy3DLa:becwvh) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/e770aadb7869b677dba0c5ee4c6c37964b1fe36b/integration_tests/devices/payload.py#L22)):*

It will be better to provide callback methods returning processing time (or no time at all) 
</blockquote></details>

Added callback function returning time. No function set is timeout 0.

---

*[integration_tests/devices/payload.py, line 195 at r3](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197#-Ko8aY1w2I1R35GPG_os:-KoLiF2zFYu69DFjc0fi:b7xku50) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/e770aadb7869b677dba0c5ee4c6c37964b1fe36b/integration_tests/devices/payload.py#L195)):*

By default all calls to I2C Device methods are logged - no need to log them second time
</blockquote></details>

Logging removed.

---

*[libs/drivers/payload/commands/SunS.cpp, line 35 at r3](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197#-Ko8nuzvmRG9IJuRKay8:-KoLgo8MWb9dLS4DH7GK:byw4n7) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/e770aadb7869b677dba0c5ee4c6c37964b1fe36b/libs/drivers/payload/commands/SunS.cpp#L35)):*

I don't think it is necessary to log every received value
</blockquote></details>

Logging values in driver removed.

---

*[libs/drivers/payload/commands/Whoami.cpp, line 17 at r3](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197#-Ko8nKG8frN2GANxt6Ok:-KoLgvirRhgYg5kWilXO:b-9oyzuf) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/e770aadb7869b677dba0c5ee4c6c37964b1fe36b/libs/drivers/payload/commands/Whoami.cpp#L17)):*

We tend to use `[module] Message` convention for log message - just a suggestion
</blockquote></details>

Changed logging messages to include [Payload] tag

---

*[libs/drivers/payload/Include/payload/payload.h, line 97 at r3](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197#-Ko8pTBNgJ_kHFZ7PuQL:-KoLh-pO8bXZ4f7Qryvr:bsay9yv) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/e770aadb7869b677dba0c5ee4c6c37964b1fe36b/libs/drivers/payload/Include/payload/payload.h#L97)):*

30 minutes is quite a long time...
</blockquote></details>

Reduced timeout to 5 minutes (5*expected RadFET measurement).

---

*[libs/drivers/payload/Include/payload/telemetry.h, line 87 at r3](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197#-Ko8p5re-ruo2MBw0Lin:-KoLhVPoidOTlEIVoJsk:bt2vr1b) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/e770aadb7869b677dba0c5ee4c6c37964b1fe36b/libs/drivers/payload/Include/payload/telemetry.h#L87)):*

You can streamline form of this code a bit by defining base struct taking address and length as template parameters and exposing them as constexprs
</blockquote></details>

Fixed.

---

*[libs/drivers/payload/Include/payload/telemetry.h, line 103 at r3](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197#-Ko8oyAF-gmJ0veMjUlz:-KoLh9geB09jVjFGwie7:ba3dfob) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/e770aadb7869b677dba0c5ee4c6c37964b1fe36b/libs/drivers/payload/Include/payload/telemetry.h#L103)):*

As you are no longer relaying on memory layout you can remove this assert
</blockquote></details>

Removed

---

*[libs/drivers/payload/Include/payload/commands/base.h, line 75 at r3](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197#-Ko8oWctmg_LkWWCvqK4:-KoLhXftT6pONWErsd7j:b-886gsx) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/e770aadb7869b677dba0c5ee4c6c37964b1fe36b/libs/drivers/payload/Include/payload/commands/base.h#L75)):*

I think it is possible to remove almost all virtual calls by using constexprs inside command classes (for data address and buffer). Most of the code in commands is the same
</blockquote></details>

Fixed

---

*[libs/drivers/payload/Include/payload/commands/base_code.hpp, line 86 at r3](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197#-Ko8mS7ClTNB7qSZODDH:-KoLhbnAzL-C4rK9fHAu:b-w6mnqz) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/e770aadb7869b677dba0c5ee4c6c37964b1fe36b/libs/drivers/payload/Include/payload/commands/base_code.hpp#L86)):*

why splitting into two files?
</blockquote></details>

I wanted to separate declarations from "definitions" and reduce adding unnecessary includes to other areas. I can merge those two files if it would be better.

---

*[platforms/FlightModel/Include/io_map.h, line 32 at r3](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197#-Ko8ifV6Q57x1D5MU5Ss:-KoLhtoG_dKn3au46J5h:b-yjckyk) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/e770aadb7869b677dba0c5ee4c6c37964b1fe36b/platforms/FlightModel/Include/io_map.h#L32)):*

During last tests we found out that different pin is used for payload interrupt - verify with @ggajoch and @begreeen 
</blockquote></details>

Discussed and I will update it after new platform code will be ready

---

*[src/commands/payload.cpp, line 233 at r3](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197#-Ko8jXIRlkkOqhDRFpsu:-KoLi0UR9H2GiHX3Pt0Q:bssorlh) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/e770aadb7869b677dba0c5ee4c6c37964b1fe36b/src/commands/payload.cpp#L233)):*

In all commands: provide labels for each value in output. It's much easier to understand (and determine if it's correct)
</blockquote></details>

Labels for values added

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197)*

Novakov:
Reviewed 30 of 30 files at r4.
Review status: all files reviewed at latest revision, 11 unresolved discussions.

---

*[integration_tests/tests/test_payload.py, line 15 at r4](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197#-KoMVhZXTgQevdB6K4Oz:-KoMVhZXTgQevdB6K4P-:bp0s4bc) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/95dde43276a8cb737c75bcee1be731c2660b2ea5/integration_tests/tests/test_payload.py#L15)):*
> ```Python
> 
>     def decode_result(self, result_string):
> ```

empty lines?

---

*[integration_tests/tests/test_payload.py, line 104 at r4](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197#-KoMVpbCkwQM_Yg1E-I1:-KoMVpbCkwQM_Yg1E-I2:b9r4cge) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/95dde43276a8cb737c75bcee1be731c2660b2ea5/integration_tests/tests/test_payload.py#L104)):*
> ```Python
> 
>         self.system.payload.set_timeout_callback()
>         result = self.system.obc.payload_radfet_off()
> ```

You can move decoding results to decorator on methods in OBC mixin - that way it will work also in pyconsole

---

*[libs/drivers/payload/Include/payload/commands/Housekeeping.h, line 19 at r4](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197#-KoMW5_EYMhvvz1TnGNl:-KoMW5_EYMhvvz1TnGNm:b-gytqa4) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/95dde43276a8cb737c75bcee1be731c2660b2ea5/libs/drivers/payload/Include/payload/commands/Housekeeping.h#L19)):*
> ```objc
>                                             PayloadTelemetry::Housekeeping,
>                                             PayloadTelemetry::Housekeeping::DeviceDataAddress,
>                                             PayloadTelemetry::Housekeeping::DeviceDataLength>
> ```

If we assume (I think we can) that all telemetry classes will have members `DeviceDataAddress` and `DeviceDataLength` than there is no need to pass them by template parameters.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197)*

tom4r:
Review status: 46 of 57 files reviewed at latest revision, 11 unresolved discussions.

---

*[integration_tests/tests/test_payload.py, line 15 at r4](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197#-KoMVhZXTgQevdB6K4Oz:-KoNy-EJqMcbXaiNmaqU:bt2vr1b) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/95dde43276a8cb737c75bcee1be731c2660b2ea5/integration_tests/tests/test_payload.py#L15)):*

empty lines?
</blockquote></details>

Fixed.

---

*[integration_tests/tests/test_payload.py, line 104 at r4](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197#-KoMVpbCkwQM_Yg1E-I1:-KoNy0bmcKg-7gEfbo7b:b-896fix) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/95dde43276a8cb737c75bcee1be731c2660b2ea5/integration_tests/tests/test_payload.py#L104)):*

You can move decoding results to decorator on methods in OBC mixin - that way it will work also in pyconsole
</blockquote></details>

Done.

---

*[libs/drivers/payload/Include/payload/commands/Housekeeping.h, line 19 at r4](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197#-KoMW5_EYMhvvz1TnGNl:-KoNxwSMYHzCjR4DtpZs:b-9qjhto) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/95dde43276a8cb737c75bcee1be731c2660b2ea5/libs/drivers/payload/Include/payload/commands/Housekeeping.h#L19)):*

If we assume (I think we can) that all telemetry classes will have members `DeviceDataAddress` and `DeviceDataLength` than there is no need to pass them by template parameters.
</blockquote></details>

Thank you. Removed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197)*

Novakov:
Reviewed 11 of 11 files at r5.
Review status: all files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197)*

Novakov:
Reviewed 13 of 14 files at r6.
Review status: all files reviewed at latest revision, 6 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197)*

zaq32:
Reviewed 5 of 30 files at r2, 8 of 30 files at r4, 3 of 11 files at r5, 6 of 14 files at r6.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[integration_tests/devices/payload.py, line 9 at r6](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197#-Komo9FK6cIeSvXwKIWq:-Komo9FK6cIeSvXwKIWr:b-yffoy1) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/ab39dbc08ba227f353ae2f4d413a589d01a2e340/integration_tests/devices/payload.py#L9)):*
> ```Python
> 
> 
> class Payload(I2CDevice):
> ```

it would be usefull to have a way of hooking into the mock to be able to customize its behaviour in tests

---

*[libs/drivers/payload/Include/payload/commands/base.h, line 74 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197#-Kns_BazpAPTOlZDeu4M:-KomnneH9suF5TaBdSPT:b69vqo6) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/3095f42da92104d1b02b32667aa3a55521326026/libs/drivers/payload/Include/payload/commands/base.h#L74)):*
<details><summary><i>Previously, tom4r wrote…</i></summary><blockquote>

This field is protected.
</blockquote></details>

Protected field is just as bad as public one. The only difference is the length of the list of its potential victims

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197)*

zaq32:
Reviewed 1 of 53 files at r1, 6 of 30 files at r2, 6 of 30 files at r4, 5 of 11 files at r5, 7 of 14 files at r6.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197)*

tom4r:
Review status: 52 of 57 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[integration_tests/devices/payload.py, line 9 at r6](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197#-Komo9FK6cIeSvXwKIWq:-KovCerBnuVgrextwth7:bsoy3r2) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/ab39dbc08ba227f353ae2f4d413a589d01a2e340/integration_tests/devices/payload.py#L9)):*

it would be usefull to have a way of hooking into the mock to be able to customize its behaviour in tests
</blockquote></details>

Rewritten driver to allow overriding payload processing mocks and changed internal values from magical lists to proper classes.

---

*[libs/drivers/payload/Include/payload/commands/base.h, line 74 at r1](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197#-Kns_BazpAPTOlZDeu4M:-KovD6DgHOLxKfr1oT8k:b-wasczs) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/3095f42da92104d1b02b32667aa3a55521326026/libs/drivers/payload/Include/payload/commands/base.h#L74)):*

Protected field is just as bad as public one. The only difference is the length of the list of its potential victims
</blockquote></details>

You are right, fixed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197)*

zaq32:
Reviewed 1 of 53 files at r1, 1 of 30 files at r2, 11 of 12 files at r7.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197)*

zaq32:
Reviewed 4 of 5 files at r8.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/197)*

PR #236, opened 2017-08-04
Telemetry imtq

zaq32:
Extend telemetry acquisition with imtq state.

The imtq telemetry collection is collected as the imtq driver is being used by different obc elements (mainly adcs plugins and/or experiments).  Mainly because of the following reasons:
* adcs experimens (sunpointing/detumbling) are not very tolerant for unexpected unbounded delays in the middle that could be caused by the asynchronous process trying to collect telemetry. 
* it is difficult to collect imtq magnetometer measurements, b-dot & dipoles without knowing exact current imtq (adcs) operating mode. In case of imtq built-in detumbling they are acquired using `get detumble data`, however in case of experimental detumbling/sunpointing `get detumble data` cannot be used and those telemetry parts have to be acquired using other means.

Based on that the proxy layer for imtq drver has been implemented that captures the necessary telemetry elements as they are available and stored in the temporary storage.  The actual acquisition element only captures the telemetry from that temporary storage.

Additional small changes:
* Replaced several small telemetry elements with single template based implementation & some typedefs with tags.

---
This change is [Reviewable](https://reviewable.io/reviews/pw-sat2/pwsat2obc/236)


zaq32:
[[Telemetry] Add mechanism of gathering the telemetry from obc components](https://trello.com/c/Eylh5zUA/165-telemetry-add-mechanism-of-gathering-the-telemetry-from-obc-components)

Novakov:
Reviewed 41 of 46 files at r1, 11 of 11 files at r2.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[libs/telemetry/imtq/ImtqTelemetryCollector.cpp, line 12 at r2](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/236#-Kr1Pw8RRax19pUhlMGz:-Kr1Pw8RRax19pUhlMH-:b-6lnwof) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/ec6f178e5210c5058d7396d976487c0ab9c10767/libs/telemetry/imtq/ImtqTelemetryCollector.cpp#L12)):*
> ```cpp
>         ImtqTelemetryCollector::ImtqTelemetryCollector(IImtqDriver& driver) : next(driver), semaphore(nullptr)
>         {
>             std::uninitialized_fill(elementUpdated.begin(), elementUpdated.end(), false);
> ```

why not `elementUpdated.fill(false)`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/236)*

zaq32:
Review status: 42 of 50 files reviewed at latest revision, 1 unresolved discussion.

---

*[libs/telemetry/imtq/ImtqTelemetryCollector.cpp, line 12 at r2](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/236#-Kr1Pw8RRax19pUhlMGz:-Kr4Q9Uq-SE8pVyv57oY:b7a0zgc) ([raw file](https://github.com/pw-sat2/pwsat2obc/blob/ec6f178e5210c5058d7396d976487c0ab9c10767/libs/telemetry/imtq/ImtqTelemetryCollector.cpp#L12)):*

why not `elementUpdated.fill(false)`?
</blockquote></details>

in this case both will work just fine.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/236)*

Novakov:
Reviewed 8 of 8 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/236)*

PR #252, opened 2017-08-14
EPS HK/UPTIME printf specifier fix in OBC terminal

pkuligowski:
0xFFFF was printed as -1 in OBC terminal (long uint conversion to long int). This number should be printed as 4294967295. The %lu instead of %ld printf specifier fixes this issue.

---
This change is [Reviewable](https://reviewable.io/reviews/pw-sat2/pwsat2obc/252)


zaq32:
Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/252)*

PR #361, opened 2017-10-17
Delay after turning on overheat mode

ggajoch:
If `IgnoreOverheat` and `EnableMainThermalKnife` are send simultaneously there is a possiblity of race condition when:
  - overheat was detected, so turning off LCLs
  - in the middle, interrupt is server to disable overheat mode (but on-going check is still going)
  - then, before end, TK is enabled and immediatelly disabled by overheat protection.

This event is very unlikely (there is a window of about 2ms each second, and those two command would have to be send with less than 2ms interval. However fix was simple and should be enough.

---
This change is [Reviewable](https://reviewable.io/reviews/pw-sat2/pwsat2obc/361)

### DIFF
--- a/libs/mission/sail/Include/mission/sail.hpp
+++ b/libs/mission/sail/Include/mission/sail.hpp
@@ -179,7 +179,7 @@ namespace mission
         using StepProc = void (*)(OpenSailTask* This, SystemState& state);
 
         /** @brief Sail opening steps */
-        static StepProc Steps[24];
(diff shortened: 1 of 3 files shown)


Novakov:
Reviewed 3 of 3 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pw-sat2/pwsat2obc/361)*

